PR #1, opened 2015-07-10
MediaRanker complete

blip-lorist:
I'm very glad this was an exploratory project - I learned a TON from it! Stuck with the single model structure through the duration of the project, and I was able to remove excess controllers and shrink it down to just one controller. Very pleased with that. Thank you instructors for testing this app and trying to break it!

+1 for Bootstrap. I feel like this tool will save us SO much time with styling apps.


Hamled:
First off, I love how consolidated all the code is in this project! With the singular model class, simple controllers, and heavy use of partials this is a very lean (and thus more easily specified and understood) codebase.

blip-lorist:
Mahalo to @Hamled for the really helpful feedback! Got rid of the params[:format] thing and updated it to params[:media_type] - at the time of this project, I didn't understand what [:format] stored, so this info was really helpful in giving me some things to research.  

I've implemented all suggestions mentioned here in my latest pushed commits. Cheers!

Hamled:
Awesome! Glad to know the feedback has been helpful @lorainekv :)

PR #2, opened 2015-07-10
Complete MediaRanker

acmei:
I tried to use partials (for views) and shared examples (for specs) where I could to DRY out my code as much as possible. I learned a lot more about testing in Rails with this project. Also, since we utilized bootstrap and were instructed to design our website similar to an example, I felt like it saved me from having to make CSS design decisions. I was proud of how I was able to incorporate the shared code among all of my pages and specs and deleted tons of lines of code.

**Things I Learned via MediaRanker**  
- Rails rspec testing, creating and writing specs for models and controllers
- rspec shared examples
- Bootstrap
- polymorphic paths
- model validations
- scopes

Thanks for your feedback!


app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,62 @@
+class AlbumsController < ApplicationController
+
+  before_action :get_album, only: [:show, :edit, :update, :upvote, :destroy]
+
+  def get_album
```
GreyKn: Considering this is an internal only call, consider moving it to private

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,62 @@
+class AlbumsController < ApplicationController
+
+  before_action :get_album, only: [:show, :edit, :update, :upvote, :destroy]
```
GreyKn: If you have 7 actions, and 5 require this before_action, consider using `except:`, that way you can list less items

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+  def new
+    @album = Album.new
+  end
+
+  def create
+    @album = Album.new(album_params)
+    if @album.save
+      redirect_to album_path(@album.id)
+    else
+      render :new
```
GreyKn, line 22: While this works, if the album save fails, it renders the new page template with no context on why the user is presented with the `new` template.

Considering creating a modified new template contains an error message for create's else code path to use.

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    redirect_to album_path
+  end
+
+  # Upvote an album
+  def upvote
+    @album.vote += 1
+    @album.save
+
+    redirect_to :back rescue redirect_to album_path
```
GreyKn, line 40: Avoid doing blind rescues when possible. Catching the specific error you're expecting to be thrown is often best practice, as catching all errors can hide other errors that are happening.

app/views/albums/new.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= render partial: "shared/form", locals: { title: "New Album",
```
GreyKn, line 1: I notice that your `edit` and `new` erb templates are the same. Any reason you can't use a single one?

app/views/medium/index.html.erb
```diff
@@ -0,0 +1,23 @@
+<div class="col-md-4">
+  <h1>Top Books</h1>
+  <% @books.each do |book| %>
```
GreyKn, line 3: Good use of enumerables in your template :+1:

spec/support/shared_controller_examples.rb
```diff
@@ -0,0 +1,129 @@
+RSpec.shared_examples "MediaController" do
```
GreyKn, line 1: :heart:  shared examples

spec/support/shared_controller_examples.rb
```diff
@@ -0,0 +1,129 @@
...
+        it "creates a new instance of Media" do
+          get :new, media_params
+          expect(assigns(media)).to be_kind_of(model)
+        end
+      end
+
+      context "invalid params" do
+        it "doesn't instantiate Media with invalid params" do
+          get :new, invalid_params
+          expect(model.count).to eq 0
```
GreyKn, line 29: Wouldn't the model count still be 0 even if a new Media item was instantiated in this fashion?

spec/support/shared_controller_examples.rb
```diff
@@ -0,0 +1,129 @@
...
+        it "renders the :new view (to allow users to fix invalid data)" do
+          post :create, invalid_params
+          expect(response).to render_template(:new)
+        end
+      end
+    end
+
+    describe "PATCH #update" do
+      before :each do 
+        @created = model.create(name: "name")
```
GreyKn, line 62: While this works for this test, something to consider is that every time you `create` something for a test, you'll want to create a subsequent `destroy` as well.

So you could add something like this:

```
after :each do
  @created.destroy
end
```

spec/support/shared_controller_examples.rb
```diff
@@ -0,0 +1,129 @@
...
+      
+      context "valid params" do
+        it "updates an existing record" do
+          patch :update, { id: @created.id }.merge(new_params)
+          # NOTES TO REMEMBER:
+          # put :update, id: @created.id, { book: { name: "title" } }
+          # put :update, { id: @created.id, book: { name: "title" } }
+          # watch out for implicit parens!!!
+          @created.reload
+          expect(@created.name).to eq "title"
```
GreyKn, line 73: `"title"` is sort of magical, so considering using something like 

```
expected(@created.name).to eq new_params[:book][:name]
```

That way, if you ever have to change the name for whatever reason, you don't have to rewrite the test.

GreyKn:
Hey Anita,

Good on this project. I liked that you used shared examples to DRY up your tests. Looking at that, do you think could could dry up your controllers as well?

I left a couple of comments about how you might make your tests more resilient.

Let me know if you have any questions.

spec/support/shared_controller_examples.rb
acmei, line 62: Hey @GreyKn, question about this comment. I thought the before :each action in the testing environment creates the item for that specific test and then deletes it once the test has run...is this inaccurate?

acmei:
@GreyKn I definitely see that my controllers can be DRYed out since they're mostly using the same methods. I looked into this and don't really understand the best way to go about doing this. I read about concerns and super controllers. What would you recommend?

spec/support/shared_controller_examples.rb
GreyKn, line 62: The `:each` says "before each `it`, do this thing". It never deletes it anything explicitly. 

Usually what happens is that you assign a variable, which goes out of scope beyond the test, thus "deleting" it. 

But when you use persistent stores like a database, the last thing you create may stick around and mess with your tests.

GreyKn:
There are a couple ways of doing it. I don't think you covered modules so I won't go into that here, but you could have written all the methods generically inside application_controller, then set the specific media types (like models) inside the specific controller.

So like

```
class ApplicationController < ActionController::Base
  before_action :get_media, except: [:index, :new, :create]
  def new
    @media = self.model.new
  end

  def upvote
    @media.vote += 1
    @media.save
    redirect_to :back rescue redirect_to self.media_path    
  end

private
  def get_media
    @media = self.model.find(params[:id])
  end
end
```

```
class AlbumController < ApplicationController
  def self.model
    Album
  end

  def self.media_path
    album_path
  end

private
  def media_params
    params.require(:album).permit(:name, :artist, :desc)
  end
end
```

```
class BookController < ApplicationController
  def self.model
    Book
  end

  def self.media_path
    book_path
  end

private
  def media_params
    params.require(:book).permit(:name, :author, :desc)
  end
end
```

acmei:
We have talked about modules. It's been a while though. I'll try this implementation. But if you have time would you mind explaining how this could be done with modules? Thanks!

PR #3, opened 2015-07-10
features: a single controller, two tables in the DB, & custom URL helpers

sojeri:
I really enjoyed this project and would like some extra feedback on my specs.
- Where have I done too much or not enough?
- What are some better ways to test these conditions?
- Is there anything I completely missed?

Well, it's probably also useful to know that for the project as a whole. >_>

Also, sometimes when running `db:reset` or `db:migrate` I had to comment out my routes to get the command to work. And sometimes I _didn't_. Watch out for that?
- I'd also love to know _why_, so if you know that or a resource I can explore please share that with me.

Thank you for reviewing my code! I super appreciate all the time it takes to read through a bunch of code and comment on best practices, etc. n_n


Hamled:
> Also, sometimes when running db:reset or db:migrate I had to comment out my routes to get the command to work. And sometimes I didn't. Watch out for that?

Do you recall which routes needed to be commented out? It'd also be helpful to know what error messages you were receiving and maybe what commit you were on so that we could attempt to reproduce it.

sojeri:
re:

> Do you recall which routes needed to be commented out? It'd also be helpful to know what error
> messages you were receiving and maybe what commit you were on so that we could attempt to
> reproduce it.

I had just been commenting out the whole file. [When I switched back to constraining the routes with regex built from a static array instead of from my categories table in the database, the issue disappeared.](https://github.com/drvonnjerryxlii/C3Projects--MediaRanker/commit/1ed2fcfbfe9aa71d99c66f5bd8ab9b27ff659f03)

Hamled:
re:

> I had just been commenting out the whole file. When I switched back to constraining the routes with regex built from a static array instead of from my categories table in the database, the issue disappeared.

My best guess is that using the `Category.all` method in your routes was the culprit. It seems likely that the routing code is loaded before any of the ActiveRecord database stuff. So, when you're attempting to `migrate` or `reset` from an empty database, it's failing to find any table associated with the `Category` model because it has not yet been created.

In general I'd say the routes configuration should avoid relying upon anything in the database, both for this particular concern, and because it makes the "routing layer" more complicated and harder to test / verify.

sojeri:
@Hamled, I just want to say again: THANK YOU! Your feedback is super helpful & I feel like I'm learning so much from it. :)

I'm merging this now so I can finish my homework for this weekend, which is a separate PR for a refactoring of this project. n_n

PR #5, opened 2015-07-10
Media Ranker final

lilagrc:
I (mostly) enjoyed this project! I felt like I really nailed partials for my views, and I LOVE them. Along with some others in class, I found polymorphic paths and they helped me get the partials cleaned up. I'm especially happy with my welcome index page. I also really enjoyed using bootstrap and it was nice to be able to spend my time on other things and worry less about design decisions. 

I think my main weak point for this project was on rspec tests. I had written most of my model tests and was tired of all the copy and pasting, so I spent time working on shared examples. These are also magic, but I really struggled getting them to work for my controller tests. I managed to get shared examples for about half of my controller tests, but then started to run out of time. Focusing on partials and style I think came at the detriment of writing better tests, so I think most of them are pretty meh. However, some other Adies found solutions that came out during the project debrief, so with more time, I feel confident I'd be able to have all of my tests use shared examples.


app/assets/images/Media Ranker.html
```diff
@@ -0,0 +1,53 @@
...
+<a class="btn btn-danger" data-confirm="Are you sure you want to delete this?" rel="nofollow" data-method="delete" href="./Media Ranker_files/Media Ranker.html">Delete</a>
+
+
+
+
+
+  </div>
+
+
+</body></html>
```
jnf: I'm confused. What's this file and why's it in `app/assets/images`?

app/assets/stylesheets/application.scss
```diff
@@ -0,0 +1,33 @@
...
+}
+
+html {
+  margin-right: 76px;
+  margin-left: 76px;
+}
+
+
+ @import "bootstrap-sprockets";
+ @import "bootstrap";
```
jnf, line 29: Generally it's better to put `@import` commands before any specific styling. Because CSS is procedural, the specific rules you apply to `.page-header` and the `html` elements could be overwritten by what's in the `@import` files.

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,64 @@
+class AlbumsController < ApplicationController
+
+  def self.model
+    Album
```
jnf, line 4: For future reference, you can ask any Ruby object it's class by just calling `object.class`.

``` ruby
"test".class # => String
"test".class == String # => true
```

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+  def show
+    @album = Album.find(params[:id])
+  end
+
+  def upvote
+    album = Album.find(params[:id])
+    album.add_vote
+    album.save
+
+    redirect_to album_path
```
jnf, line 22: I think this redirect wouldn't work without passing the album to the path, like you're doing on line 34 below (`redirect_to album_path(album)`).

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    if @album.save
+      redirect_to album_path(@album)
+    else
+      render :new
+    end
+  end
+
+  def edit
+    show
```
jnf, line 41: I'm curious--why are you calling `show` here? While this is totes valid Ruby (calling a method in a method), it's not a common Rails pattern. There's nothing wrong with that, I'm just curious as to your reasoning.

lilagrc, line 41: I believe my FarMarRails partner and I did this as a way to get the an instance of a model (with the necessary params id). Since the show action sets the instance variable to @album, rather than writing that code again, I put show to do the same thing. I have no idea if this is a good pattern, or one best avoided, so any advice is welcome!

jnf, line 41: Ah! That makes sense, given what we'd learned about Rails at the time. Moving forward, we'll use [before filters](http://guides.rubyonrails.org/action_controller_overview.html#filters) for this kind of functionality.

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+  def show
+    @book = Book.find(params[:id])
+  end
+
+  def upvote
+    book = Book.find(params[:id])
+    book.add_vote
+    book.save
+
+    redirect_to book_path
```
jnf, line 22: Looking at this, my guy says this needs to be `redirect_to book_path(book)`, but I'm testing it locally (I cloned your repo), and it seems to be working. I'm kinda confused. I'm mentioning it here as a reminder to myself to do some more research. :)

app/views/albums/edit.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= render partial: 'shared/form', locals: {  object: @album,
```
jnf, line 1: Good use of `locals` here. It really helps us keep the form generic, which gives it greater reusability.

app/views/shared/_form.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= render 'shared/header' %>
+
+<h2><%="#{action} #{object.class}" %></h2>
+
+<%= form_for object, html: {class: "form"} do |f| %>
+  <% {name: :name, creator: "#{creator}", description: :description}.each do |key, value| %>
```
jnf, line 6: The interpolation in the `:creater` hash value is unnecessary. The local variable creator is already a string ("Director" or similar), so doing interpolation is essentially just calling `to_s` on something that's already a string. You'd be able to just say `{ creator: creator }` here.

app/views/shared/_index.html.erb
```diff
@@ -0,0 +1,28 @@
...
+      <td><%= link_to object.name, polymorphic_path(object) %></td>
+      <td><%= link_to "Upvote", polymorphic_path([:upvote, object]), method: :post, class: "btn btn-default" %></td>
+    </tr>
+    <% end %>
+
+
+</table>
+
+<%= link_to "View All Media", root_path, class: "btn btn-default" %>
+<%= link_to "Add #{media}", "/#{object_path}/new", class: "btn btn-default" %>
```
jnf, line 22: This looks like another opportunity to use `polymorphic_path` just like you're doing in the lines above.

spec/controllers/album_controller_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      it "renders the new action" do
+        post :create, params
+
+        expect(response).to render_template("new")
+      end
+    end
+  end
+
+  describe "PATCH update" do
+    before :each do
```
jnf, line 46: Excellent use of `before` here. You could have done this setup in the `it` block, but extracting it here means you're ready to write the next test (and there's _always_ a next test).

spec/support/media_shared_controller_spec.rb
```diff
@@ -0,0 +1,77 @@
+require 'spec_helper'
+
+RSpec.shared_examples "a media" do
+
+  let(:medium) { described_class.model.new() }
```
jnf, line 5: You set up a `let` here but I don't think you use it anywhere in the file. If it's deprecated or unused, it'd be best to delete it.

jnf:
Hey! Good job on this project. I've left a few notes that are mostly about tweaking style for clarity or efficiency. Overall, though, your code looks great. The shared examples looks pretty good and clearly demonstrate that you understand where the overlap in functionality/coverage is. :+1:

app/controllers/books_controller.rb
lilagrc, line 22: Yes, this is super weird! I tested this when you pointed it out on the album controller and it works... but it doesn't seem like it should.

PR #6, opened 2015-07-10
MediaRanker

ktrops:
I first did three models and three controllers, but then I decided to try one controller and one models. If you creep on my git repo there's a branch called kja/master and that has the threes version. 

There are a lot more test that I want to create, so I may add more later. If you have any ideas on how to simplify my code, let me know.

Thanks!


.gitignore
```diff
@@ -1,35 +1,17 @@
...
+# Ignore the default SQLite database.
+/db/*.sqlite3
+/db/*.sqlite3-journal
 
-# Ignore pow environment settings
-.powenv
+# Ignore all logfiles and tempfiles.
+/log/*
+!/log/.keep
+/tmp
```
kariabancroft, line 17: I would recommend adding any/all `.` files to your gitignore. Some examples:
.rspec
.ruby-gemset
.ruby-version
app/assets/images/.keep
app/controllers/concerns/.keep
app/mailers/.keep
app/models/.keep
app/models/concerns/.keep
lib/assets/.keep
lib/tasks/.keep
log/.keep

etc

app/controllers/media_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+  before_action :find_media
+
+  def find_media
+    @movies = Medium.sort_votes(Medium.find_movies)
+    @books = Medium.sort_votes(Medium.find_books)
+    @albums = Medium.sort_votes(Medium.find_albums)
+  end
+
+  def home
+    @movies_limit = @movies.first(4)
```
kariabancroft, line 11: I'd recommend moving this `4` out into a constant

app/controllers/media_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def destroy
+    media = Medium.find(params[:id])
+    format = media.format
+    media.destroy
+    redirect_to Medium.pick_index_path(format)
+  end
+
+  def media_params
```
kariabancroft, line 79: Add the private keyword for this method

app/controllers/media_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    @media = Medium.find(params[:id])
+  end
+
+  def edit
+    @media = Medium.find(params[:id])
+    @method = :patch
+  end
+
+  def update
+    @media = Medium.find(params[:id])
```
kariabancroft, line 54: I see this same line a few different times so I'd recommend moving this somewhere shared

app/helpers/application_helper.rb
```diff
@@ -0,0 +1,93 @@
...
+    if media.format == "book"
+      book_upvote_path(media.id)
+    elsif media.format == "album"
+      album_upvote_path(media.id)
+    else
+      movie_upvote_path(media.id)
+    end
+  end
+
+  #decides on a path for show, update, and destroy action
```
kariabancroft, line 14: This is a great place to try to explore polymorphic paths in the future

app/helpers/application_helper.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  #choses a format for the format column in the table when creating a media. 
+  def pick_format
+    if request.path.include?("books")
+      "book"
+    elsif request.path.include?("movies")
+      "movie"
+    elsif request.path.include?("albums")
+      "album"
```
kariabancroft, line 76: I see the "books", "movies" and "albums" strings everywhere so I'd strongly recommend pulling those out into constants with the thought that it the name of one of these items changed, you'd only need to change it in one place.

app/models/medium.rb
```diff
@@ -0,0 +1,66 @@
...
+  before_save :default_votes
+
+  validates :name, presence: true, uniqueness: true
+  validates :description, presence: true
+  validates :creator, presence: true
+
+  #when a new media is created it sets votes to 0
+  def default_votes
+    if self.votes.nil?
+      self.votes = 0
```
kariabancroft, line 11: Could you use a `||=` here instead?

app/models/medium.rb
```diff
@@ -0,0 +1,66 @@
...
+  def self.pick_edit_path(media)
+    if media.format == "book"
+      Rails.application.routes.url_helpers.edit_book_path(media.id)
+    elsif media.format == "album"
+      Rails.application.routes.url_helpers.edit_album_path(media.id)
+    else
+      Rails.application.routes.url_helpers.edit_movie_path(media.id)
+    end
+  end
+
```
kariabancroft, line 65: Same comment about the book, album, movie strings

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,21 @@
...
+<body>
+  <div class="container-fluid">
+      <!-- jQuery (necessary for Bootstrap's JavaScript plugins) -->
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script>
+<%= render partial: "shared/header" %>
+
+<%= yield %>
+  </div>
+</body>
+</html>
```
kariabancroft, line 21: Watch out for your indentation in this file

app/views/media/home.html.erb
```diff
@@ -0,0 +1,34 @@
...
+    <div class = "col-sm-4">
+      <h1>Books</h1>
+        <% @books_limit.each do |book| %>
+        <h2><%= link_to "#{book.name}", book_path(book.id), method: :get %></h2>
+        <p>Number of votes: <%= book.votes %></p>
+      <% end %>
+      <%= button_to "view more books", book_index_path, method: :get %>
+    </div>
+
+    <div class = "col-sm-4">
```
kariabancroft, line 22: I think you could dry this view up even more since the markup is the same across each media type even though the specific names are different.

config/routes.rb
```diff
@@ -0,0 +1,74 @@
...
+scope '/albums' do
+  resources :media, as: :album
+end
+
+scope '/movies' do
+  resources :media, as: :movie
+end
+
+
+patch "/movies/:id/upvote" => "media#upvote", as: :movie_upvote
```
kariabancroft, line 31: I would remove the extra comments below this block and then indent this section to follow the indentation you used above

spec/controllers/media_controller_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+
+
+
+
+
+
+
+
+
```
kariabancroft, line 71: Ah! Why so much whitespace?

spec/models/medium_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      expect(book).to_not be_valid
+      expect(book.errors.keys).to include(:creator)
+    end 
+  end
+
+  describe "saving and creating an instance" do
+    it "sets votes to 0 upon creation" do
+      album = Medium.create(name: "Kiss Land", description: "Debut studio album", creator: "The Weekend")
+      expect(album.votes).to eq 0
+    end
```
kariabancroft, line 29: I think you should your test suite around the `create` functionality

kariabancroft:
Nice job on this! I'm glad you attempted to use the one model/controller. I think there are still a few things that you can DRY up. I'd be cautious of using "magic strings" everywhere. If you get in the habit of using the same strings over and over again, you should consider using constants for them.

app/views/media/home.html.erb
ktrops, line 22: Maybe a partial would work here?

spec/controllers/media_controller_spec.rb
ktrops, line 71: I always add whitespace and the end if think I'm going to add more code because I don't like working at the bottom of the screen. I'll try to remember to remove it.

.gitignore
ktrops, line 17: That's a good idea! I'll definitely do that with my next projects.

app/controllers/media_controller.rb
ktrops, line 54: I wan't sure if I should have made a separate method because calling a method that I made would be the same, if not more characters.

app/helpers/application_helper.rb
ktrops, line 14: ooooo...awwwww...polymorphic paths are so cool! I definitely need to read up on them.

PR #9, opened 2015-07-10
Media Ranker Project! 

shannachau:
I really enjoyed this project. I liked copying off of something that someone else already made. It was easier to practice styling without having to worry about making actual design decisions. :) BOOTSTRAP IS THE BUSINESS!!! Looking forward to exploring it more in the future.

My biggest struggle was definitely writing rspec tests. I'm still not super comfortable with the syntax or how to write tests for certain methods/routes/etc. Most of the time, I knew what I wanted to test, but had trouble actually writing out the test. However, this time around, it was good practice to struggle with shared examples. I got some shared examples for the controllers, but not all of them. I'll definitely want to try more in future projects.

Also, I made lots and lots of commits!


app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,70 @@
+class AlbumsController < ApplicationController
+  def self.model
+    Album
+  end
+
+  def index
+    @albums = Album.all.order('ranking DESC')
```
kariabancroft, line 7: It would make more sense to put this specific logic in your model - potentially in a scope or method

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def create
+    @album = Album.create(create_params[:album])
+
+    if @album.save
```
kariabancroft, line 13: I like that you used this as your conditional

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+  def edit
+    @album = Album.find(params[:id])
+  end
+
+  def show
+    @album = Album.find(params[:id])
+  end
+
+  def update
+    @album = Album.find(params[:id])
```
kariabancroft, line 33: I see this line repeated numerous times to this would be a great place to use a before_action for these specific actions

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+  def edit
+    @book = Book.find(params[:id])
+  end
+
+  def show
+    @book = Book.find(params[:id])
+  end
+
+  def update
+    @book = Book.find(params[:id])
```
kariabancroft, line 33: Same comment re: DRYing this repeated logic

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      if @book.save
+        redirect_to book_path(@book)
+      else
+        render :edit
+      end
+
+    else
+      new_ranking = @book.ranking + 1
+      @book.update(ranking: new_ranking)
```
kariabancroft, line 53: You could also do `@book.ranking += 1` with a save afterwards

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,70 @@
+class BooksController < ApplicationController
+  def self.model
+    Book
+  end
+
+  def index
+    @books = Book.all.order('ranking DESC')
```
kariabancroft, line 7: Same comment about model logic

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,68 @@
+class MoviesController < ApplicationController
+  def self.model
+    Movie
+  end
+
+  def index
+    @movies = Movie.all.order('ranking DESC')
```
kariabancroft, line 7: Same comment about model logic

app/controllers/welcome_controller.rb
```diff
@@ -0,0 +1,7 @@
+class WelcomeController < ApplicationController
+  def index
+    @top_movies = Movie.all.top(10)
```
kariabancroft, line 3: Use a constant to represent the top values OR you could also encapsulate this logic within the model

kariabancroft, line 3: Nevermind! this is using a scope so then I'd just go with the constant suggestion

app/views/albums/edit.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= render :partial => 'shared/form',
```
kariabancroft, line 1: Excellent! I'm glad that you're using a partial here

app/views/albums/index.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= render :partial => 'shared/index',
```
kariabancroft, line 1: Excellent! I'm glad that you're using a partial here

app/views/shared/_index.html.erb
```diff
@@ -0,0 +1,20 @@
...
+        <td> <%= link_to media.name, polymorphic_path(media) %> </td>
+        <td> <%= button_to "Upvote", polymorphic_path(media), method: "patch", class: "btn btn-primary" %> </td>
+      </tr>
+    <% end %>
+  </tbody>
+</table>
+
+
+<%= link_to "View All Media", root_path, class: "btn btn-default" %>
+<%= link_to "Add a #{media_name}", new_polymorphic_path(media_name), class: "btn btn-default" %>
```
kariabancroft, line 20: AWESOME that you're using the polymorphic paths!

app/views/welcome/index.html.erb
```diff
@@ -0,0 +1,44 @@
+<div class="row">
+  <div class="col-lg-12">
+    <div class="row">
+      <div class="col-md-4">
+        <h1>Top Movies</h1>
```
kariabancroft, line 5: I think you could additional separate this view out into partials since each section is the same

db/migrate/20150706223809_create_books.rb
```diff
@@ -0,0 +1,12 @@
+class CreateBooks < ActiveRecord::Migration
+  def change
+    create_table :books do |t|
+      t.string :name
+      t.string :author
+      t.string :description
+      t.integer :ranking, :default => 0
```
kariabancroft, line 7: Nice use of the default on this

spec/controllers/albums_controller_spec.rb
```diff
@@ -0,0 +1,105 @@
+require 'rails_helper'
+
+RSpec.describe AlbumsController, type: :controller do
+  describe AlbumsController do
+    it_behaves_like "a media controller"
+  end
+
+  describe "POST #create" do
+    context "Valid album params" do
+      let(:album_params) do
```
kariabancroft, line 10: You can set up the shared controller to accept params like this so that you can use different params for each different type of controller. If you did this, then I think you could move ALL of the tests that are in each individual controller test into the shared example. To do this, you can just move your `let` statements to be within the `describe AlbumsController do` block after the `it_behaves_like`.

spec/support/media_shared_model_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  describe "top scope" do
+    before :each do
+      @media1 = described_class.create(name: 'whatever', ranking: 3)
+      @media2 = described_class.create(name: 'whatever', ranking: 1)
+      @media3 = described_class.create(name: 'whatevz',  ranking: 2)
+      @media4 = described_class.create(name: 'whatevah', ranking: 4)
+    end
+
+    it "orders movies according to ranking, highest to lowest" do
+      top_media = [@media4, @media1, @media3]
```
kariabancroft, line 24: Hmmm I'm wondering how you ended up with the array in this particular order?

kariabancroft:
Very nice job on this! Your use of partials for the index and forms is very clean. I think you could DRY up your controller methods a bit more by consolidating the find logic and moving some sorting logic into the model. I think that moving your `let` statements in your controller tests into the shared example `describe` would allow you to remove ALL duplication with your specs.

shannachau:
Thanks for the feedback!! Will keep in mind all your suggestions for refactoring!

PR #10, opened 2015-07-10
MediaRanker

amirahaile:
Full disclosure: this week was weird. I feel like I made a lot of silly mistakes. I'm also not super confident in my understanding of everything. I'd describe my knowledge as "foggy." 

I know there are a lot of places I could've DRY'd up my code, but after Thursday it was all about testing and making sure I didn't break any functionality. I'm proud of my attempt to use shared examples for RSpec and making the most of opportunities to use partials. I might come back to this to DRY it up, but no promises.


app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,60 @@
+class AlbumsController < ApplicationController
+  before_action :find_album, only: [:upvote, :show, :edit, :update, :destroy]
```
kariabancroft, line 2: I really like the way you used this `before_action` to DRY things up. I think that it should be defined as a private method.

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  def self.model
+    Album
+  end
+
+  def find_album
+    @album = Album.find(params[:id])
+  end
+
+  def index
+    @albums = Album.all.order('ranking DESC')
```
kariabancroft: I think that this logic, since it includes the DESC, would be better suited in the model rather than the controller

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def destroy
+    @album.destroy
+    redirect_to media_path
+  end
+
+  private
+
+  def create_params
```
kariabancroft: Since you're using this method to permit params for both the create and update operations, this method should be re-named appropriately

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,60 @@
+class BooksController < ApplicationController
+  before_action :find_book, only: [:upvote, :show, :edit, :update, :destroy]
+
+  def self.model
+    Book
+  end
+
+  def find_book
```
kariabancroft: Same comment re: private method

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,60 @@
+class BooksController < ApplicationController
```
kariabancroft, line 1: Same comments apply from the AlbumsController

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,60 @@
+class MoviesController < ApplicationController
```
kariabancroft, line 1: Same comments apply from the AlbumsController

app/models/album.rb
```diff
@@ -0,0 +1,4 @@
+class Album < ActiveRecord::Base
+  validates :title, presence: true, uniqueness: true
+  scope :top_5, -> { order('ranking DESC').limit(5) }
```
kariabancroft: This is an excellent use of a scope. It would be more extendable if you accepted the top value as a parameter rather than hard-coding it

app/views/albums/show.html.erb
```diff
@@ -0,0 +1 @@
+<%= render partial: "/shared/details", locals: { medium: @album } %>
```
kariabancroft, line 1: I'm glad you're using partials for all of these views

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>C3ProjectsMediaranker</title>
```
kariabancroft: It's good to get in the habit of updating the title when you work on these apps

app/views/media/index.html.erb
```diff
@@ -0,0 +1,46 @@
+<div class="row">
+  <div class="col-sm-4">
+    <h2>Top Movies</h2>
+
+    <ul class="media-list list-group">
```
kariabancroft: Since each of these sections are the same, I think it would make sense to use a partial for this index listing as well

app/views/shared/_details.html.erb
```diff
@@ -0,0 +1,41 @@
+<div class="col-sm-5 outer-wrapper">
+  <div class="inner-wrapper panel panel-default">
+    <div class="panel-heading">
+      <h4 class="panel-title"><%= "#{medium.class.name} ##{medium.id}" %></h4>
+    </div>
+
+    <div class="panel-body">
+      <div class="btn-mutators">
+        <%= link_to "Edit", "./#{params[:id]}/edit", html_options = { class: "btn btn-warning", role: "button" } %>
```
kariabancroft, line 9: This would be a good place to start tinkering with polymorphic paths!

app/views/shared/_details.html.erb
```diff
@@ -0,0 +1,41 @@
...
+                                                   data: {
+                                                    confirm: "Are you sure you want to delete '#{medium.title}'?",
+                                                    method: :delete }
+                                                  } %>
+      </div>
+
+      <h2> <%= medium.title %> </h2>
+
+      <h3>
+        <% if medium.has_attribute?(:director) %>
```
kariabancroft, line 22: I like the way that you're using the attribute logic to determine which field to use - rather than using the medium type itself

app/views/shared/_list.html.erb
```diff
@@ -0,0 +1,62 @@
...
+  <table class="table list">
+    <tr>
+      <th class="text-center">
+        Ranking
+      </th>
+      <th>
+        Title
+      </th>
+      <th>
+        <% if media.model_name.name == "Movie" %>
```
kariabancroft: You're using these "magic strings" in a few different places so this would be a good place to use a constant or variable instead

config/routes.rb
```diff
@@ -0,0 +1,63 @@
...
+  #   get 'products/:id' => 'catalog#view'
+
+  post '/movies/:id/upvote' => 'movies#upvote'
+  post '/books/:id/upvote' => 'books#upvote'
+  post '/albums/:id/upvote' => 'albums#upvote'
+
+  resources :movies, :books, :albums
+  resources :media, only: [:index, :show]
+
+  # Example of named route that can be invoked with purchase_url(id: product.id)
```
kariabancroft: Delete the default routes info when you have your custom routes set up

spec/support/medium_controller_spec.rb
```diff
@@ -0,0 +1,153 @@
+require 'spec_helper'
+
+RSpec.shared_examples "medium" do
+
+  let(:medium_model_name){ described_class.model.name.downcase }
+
+  describe "find_medium" do
+    it "locates the correct object" do
+      medium1 = described_class.model.new(id: 1, title: "a title")
```
kariabancroft: You can just use `create` where you have all of these `new`s and then cut the lines of code in half

spec/support/medium_controller_spec.rb
```diff
@@ -0,0 +1,153 @@
+require 'spec_helper'
+
+RSpec.shared_examples "medium" do
```
kariabancroft: I'd recommend modeling your shared example name off of the name of file - the more explicit name including `controller`

spec/support/medium_controller_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+
+  let(:medium_model_name){ described_class.model.name.downcase }
```
kariabancroft, line 6: I see the use of `described_class.model` everywhere so I think it would make sense to add this to your `let` to clean up the remaining code below

spec/support/medium_controller_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+
+  describe "PUT #update" do
+    before :each do
+      medium1 = described_class.model.new(id: 1, title: "a title")
+      medium1.save
+      @medium2 = described_class.model.new(id: 2, title: "b title")
+      @medium2.save
+      described_class.model.new(id: 3, title: "c title").save
+
+      @params = { medium_model_name.to_sym => { title: "If You're Reading This It's Too Late", description: "" }, id: "2" }
```
kariabancroft, line 95: I like the way you used the `to_sym` to dynamically create the `@params`

spec/support/medium_model_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'spec_helper'
+
+RSpec.shared_examples "a medium model" do
+  describe ":top_5 scope" do
+    before :each do
+      @excluded_album = described_class.create(title: "a", ranking: 1)
+      described_class.create(title: "b", ranking: 2)
```
kariabancroft: You could use a nice little loop to do this if you created an array with `['b','c','d','e','f']` and did an each loop on this

spec/support/medium_model_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    it "only selects 5 items" do
+      expect(described_class.top_5.count).to eq 5
+      expect(described_class.top_5).to_not include(@excluded_album)
+    end
+
+    it "orders the items from high to low ranking" do
+      top_5_titles = described_class.top_5.map { |described_class| described_class.title }
+      correct_order = ["d", "e", "f", "c", "b"]
+
+      expect(top_5_titles).to eq(correct_order)
```
kariabancroft, line 25: I like the way you've created this test

kariabancroft:
Very nice job on this. I think there are a few very minor things to clean up within the controller. The partial views are really nice! You can clean up your shared examples a little bit - but overall I really like the way that you were able to include all of the necessary logic for controllers and models in them.

amirahaile:
Hey Kari! I implemented all of your suggestions (except the polymorphic path – I wasn't up for dealing with routes being finnicky on me). It's crazy how much more confident I am in Rails now compared to when I submitted this project.

PR #12, opened 2015-07-11
Completed MediaRanker project

lo-k:
Here is my completed MediaRanker project. I really enjoyed getting to play around with Bootstrap since it is so commonly talked out in the industry. I am still not the most comfortable with Rspec testing in Rails, but made some progress with it here.

Thank you for your feedback!


.gitignore
```diff
@@ -1,35 +1,19 @@
...
+/db/*.sqlite3
+/db/*.sqlite3-journal
 
-# if using bower-rails ignore default bower_components path bower.json files
-/vendor/assets/bower_components
-*.bowerrc
-bower.json
+# Ignore all logfiles and tempfiles.
+/log/*
+!/log/.keep
```
kariabancroft, line 16: Recommend adding the rest of the .keep files to your gitignore

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+    @album.save
+
+    redirect_to album_path(params[:id])
+  end
+
+
+##################### PRIVATE METHODS #####################
+  private
+
+  def create_params
```
kariabancroft, line 57: You are using this method for more than just creation so I'd recommend renaming it accordingly

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  def update
+    @album = Album.find(params[:id])
+    @album.update(create_params[:album])
+
+    redirect_to album_path(params[:id])
+  end
+
+  def destroy
+    @album = Album.find(params[:id])
```
kariabancroft, line 39: You use this find line in many different places, I'd recommend DRYing this up with a before_action that applies to these specific methods

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def new
+    @album = Album.new
+  end
+
+  def create
+    params = create_params[:album]
+    params[:rank] = 0
```
kariabancroft, line 17: I'd recommend setting this default rank within your model so that you don't need to do that here in the controller

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,60 @@
+class BooksController < ApplicationController
```
kariabancroft, line 1: Same comments from the Album controller apply to this controller

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,66 @@
+class MoviesController < ApplicationController
```
kariabancroft, line 1: Same comments from the Album controller apply here

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+
+# before_action :find_movie, only: [:update, :show, :destroy, :edit]
```
kariabancroft, line 3: You _almost_ used it! :-)

app/controllers/welcome_controller.rb
```diff
@@ -0,0 +1,9 @@
+class WelcomeController < ApplicationController
+
+  def index
+    @movies  = Movie.best(5)
```
kariabancroft, line 4: I'd recommend using a constant for this value as opposed to hard-coding it to `5`

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>C3ProjectsMediaranker</title>
```
kariabancroft, line 4: I'd make it a habit to rename the default title of your app every time

app/views/shared/_form.html.erb
```diff
@@ -0,0 +1,25 @@
+<div class="row">
+  <h1><%= verb %> <%= media_type %></h1>
```
kariabancroft, line 2: I really like the way you've been able to generalize the verb and media type necessary for the shared form

app/views/shared/_index.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  <thead>
+    <th>Ranking</th>
+    <th>Name</th>
+    <th>Upvote</th>
+  </thead>
+  <tbody>
+  <% media.each do |m| %>
+    <tr>
+      <td>Ranked: <%= m.rank %></td>
+      <td><%= link_to "#{m.name}", send(mediapath, m) %></td>
```
kariabancroft, line 11: Nice job getting these to act as routes!

config/routes.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  patch 'movies/:id/upvote' =>   'movies#upvote', as: "movie_upvote"
+  patch 'books/:id/upvote' =>    'books#upvote',  as: "book_upvote"
+  patch 'albums/:id/upvote' =>   'albums#upvote', as: "album_upvote"
+
+  resources :movies
+  resources :books
+  resources :albums
+
+  # Example resource route with options:
```
kariabancroft, line 25: Once you feel good about your routes you should delete the routes that are in the comments by default

.gitignore
kariabancroft, line 16: Also, add all of the coverage stuff to your gitignore as well

test/controllers/welcome_controller_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class WelcomeControllerTest < ActionController::TestCase
```
kariabancroft, line 3: Were all of these files in the `test` directory created in error? It seems like they should be deleted

spec/models/movie_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+        expect(movie).to_not be_valid 
+        expect(movie.errors.keys).to include(:name)
+    end
+
+    it "requires a description, all the time" do
+        movie = Movie.new
+
+        expect(movie).to_not be_valid 
+        expect(movie.errors.keys).to include(:description)
```
kariabancroft, line 16: Both of these tests include the same setup so I think it would make more sense to use a before_each that did the "newing". Then only one test would really need to the test the validity. You could also include both name and description within the same test, since both of those properties lead to a model being valid or invalid.

spec/models/album_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
+
+RSpec.describe Album, type: :model do
+  describe "model validation" do
```
kariabancroft, line 4: These first few basic tests would be a great place to try out a shared example! I'd love to see you attempt to refactor these to use one. You can utilize a `let` statement that would assign each specific model type.

spec/controllers/welcome_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  describe "GET index" do
+
+    it "renders the index template" do
+      get :index
+
+      expect(response).to have_http_status(200)
+      expect(response).to render_template("index")
+    end
+
+    it "assigns @movies to top 5 ranked movies in the database" do
```
kariabancroft, line 14: This test would be a great placed to attempt a slight refactor using some of the factorygirl functionality, especially since the specific data doesn't matter, only the count.

app/controllers/welcome_controller.rb
```diff
@@ -0,0 +1,9 @@
+class WelcomeController < ApplicationController
```
kariabancroft, line 1: I really like that you used a welcome controller to handle the index page

spec/controllers/movies_controller_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+    it "renders the index template" do
+      get :index
+
+      expect(response).to have_http_status(200)
+      expect(response).to render_template("index")
+    end
+  end
+
+  describe "GET #show" do
+    let(:movie_params) do
```
kariabancroft, line 15: I think it would ultimately be good to try a shared example for these tests. It looks like you'd need a let statement for your params hash as well as the instance of your specific model. These two let statements would be included in _this_ spec file, and then you'd create the shared example which accepts those two more generic params.

kariabancroft:
You did a very nice on this project! There were only a few cleanup items that I made note of in the controllers. I think the model specs would be a really good starting point to build up confidence in shared examples, and then you could take the leap to the shared examples with the controller methods. If you need/want any assistance with this we'd be glad to help!

.gitignore
lo-k, line 16: @kariabancroft -- Which other .keep files are you referring to? That was a default line of code in the .gitignore file and I don't see any other .keep files in the directory. Also, I had 'coverage/' written on line 19 of the .gitignore. Does that not include all of the coverage stuff?

PR #13, opened 2015-07-11
PR for media ranker

EliaMG:
## Media ranker!

I stuck fairly closely to the style of the heroku app, but I changed a few things like I wanted the whole owl to show and the display of the three lists on the home page (because it was waaay too big imo)
## Things I like: 
- bootstrap, generally, though I want to learn more about its customization and also hiding elements in resizing
- using polymorphic pathing for my routing allowed me to make all my views partials and not pass in a TON of locals, so that felt very efficient
-  feel like I managed to complete all the requirements
## Things I thought about doing (had I had unlimited time haha):

-wanted to explore using different fonts/colors in bootstrap (sure there will be more opps)
-using shared examples for tests
-refactor models to inherit from a parent Media class since all the scopes/class methods were the same


app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    @medium = Album.new
+  end
+
+  def show
+    @media = Album.find(params[:id])
+  end
+
+    def new
+    @media = Album.new
+    end
```
GreyKn, line 15: Try to keep the indentation levels consistent

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def create
+    @media = Album.new(create_params[:album])
+    @media.save
```
GreyKn, line 19: Curious, why did you use a `#new` + `#save` instead of a single `#create`?

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,53 @@
+class AlbumsController < ApplicationController
+
+  def index
+    @media = Album.all.order(rank: :desc)
+    # defining medium allows me to use polymorphic paths for new from the index page
```
GreyKn, line 5: Good comment. :+1: 

Otherwise, it could be confusing why you're defining a model instance here.

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    redirect_to album_path(@media)
+  end
+
+  def edit
+    @media = Album.find(params[:id])
+  end
+
+  def update
+    @media = Album.find(params[:id])
+    @media.update(create_params[:album])
```
GreyKn, line 29: Ah, be careful here. You're assuming that `Album.find` will find an item. There may be cases where `Album.find` will not return a record, and calling `#update` may cause weird side effects

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,52 @@
+class BooksController < ApplicationController
```
GreyKn, line 1: It looks like you're repeating a lot of code in your controllers. There might be a way to write a majority of your code in ApplicationController and just do some configuration in your specific controllers with a lot less code. Think about it.

app/views/shared/_form.html.erb
```diff
@@ -0,0 +1,25 @@
+<div class="row">
```
GreyKn, line 1: Good used of shared templates :100:

GreyKn:
Hey Elia,

Good job on this project :+1: . Great use of shared templates to drive most of your views. Having a lot of views can become an organizational nightmare of hide and seek. Think about how you could use the same principles to DRY up your controller code.

Let me know if you have any questions
-Chef

app/controllers/albums_controller.rb
EliaMG, line 29: @GreyKn, best practice on avoiding that? They would have been selecting an existing album from index or show and using a form to edit/update, so not sure how the user would be likely to be trying to update an album that didn't exist, but I could be having tunnel brain on this one. My refactor is going to involve extracting the find method into a before action however (now that I know they exist)

app/controllers/albums_controller.rb
EliaMG, line 15: Will fix in my refactor for sure!

app/controllers/albums_controller.rb
EliaMG, line 19: I'm curious too :wink:  I think my refactor will include if/else with errors and rerender, so maybe I was prematurely planning for that?

app/controllers/albums_controller.rb
GreyKn, line 29: Always assume that a record won't be found and code a path around it.

ActionControllers have a neat helper called `rescue_from` you can use to handle this:

```
class AlbumsController < ActionController::Base
  rescue_from ActiveRecord::RecordNotFound, with: some_error_handler_here
...
end
```

PR #14, opened 2015-07-11
MediaRanker - completed wave 3

vikshab:
I did some refactoring on duplicated logic, DRYed my code 

I haven't used shared_examples to test controllers. All tests are passed and I have 100% coverage. Still feels like I might have missed something to test. Would like to have comments on my partials for edit and new forms, particularly how to make less locals, the lines looks long because of them


.gitignore
```diff
@@ -1,35 +1,17 @@
...
-bower.json
+# Ignore the default SQLite database.
+/db/*.sqlite3
+/db/*.sqlite3-journal
 
-# Ignore pow environment settings
-.powenv
+# Ignore all logfiles and tempfiles.
+/log/*
+!/log/.keep
```
kariabancroft, line 16: Recommend adding all .keep files to your gitignore

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,63 @@
+class AlbumsController < ApplicationController
+
+  def index
+    @albums = Album.order('ranking DESC')
```
kariabancroft: It would be better to encapsulate the `ranking DESC` logic within the model rather than in the controller

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    if @album.save
+      redirect_to album_path(@album.id)
+    else
+      render :new
+    end
+  end
+
+  def edit
+    @album = Album.find(params[:id])
```
kariabancroft: You use this line multiple times to retrieve your model item. I'd recommend implementing a before_action that applies to each of these specific actions

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def update
+    @album = Album.find(params[:id])
+
+    @album.update(create_params[:album])
```
kariabancroft: the `update` method will actually do the save so you don't need to call `save` again

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+    @album.ranking += 1
+    @album.save
+
+    redirect_to album_path(@album.id)
+  end
+
+  # ______________________________________________________________________
+  private
+
+  def create_params
```
kariabancroft: You use this method for more than just model creation so I'd recommend re-naming this to be more generic

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    if @album.save
+      redirect_to album_path(@album.id)
+    else
+      render :edit
+    end
+  end
+
+  def destroy
+    show
```
kariabancroft: I like that you use the show logic here to retrieve your album, but this would be better suited in a before action or separate private method call because when the destroy action is called, it is not showing anything so calling this method is a bit confusing

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,63 @@
+class BooksController < ApplicationController
```
kariabancroft, line 1: Same comments from Albums controller apply here

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,63 @@
+class MoviesController < ApplicationController
```
kariabancroft, line 1: Same comments from Albums controller apply here

app/models/album.rb
```diff
@@ -0,0 +1,7 @@
+class Album < ActiveRecord::Base
+  # Validations ________________________________________________________________
+  validates :name, presence: true
+
+  # Scope_______________________________________________________________________
+  scope :best, -> { order('ranking DESC').limit(10) }
```
kariabancroft: It might be better to have this scope (and the other models) accept a parameter and then use one single constant within the UsersController index action

app/views/albums/show.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= render partial: "shared/show_content", locals: {name: @album.name, by: @album.artist, ranking: @album.ranking, description: @album.description } %>
+
+<%= render partial: "shared/buttons_footer", locals: {upvote_name_path: upvote_album_path(@album.id), edit_name: @album.name, edit_name_path: edit_album_path(@album.id), name_path: album_path(@album.id), view_name: "Albums", all_names_path: albums_path} %>
```
kariabancroft: I would recommend putting the locals on separate lines so like:

```
<%= render partial: "shared/buttons_footer", 
                       locals: {upvote_name_path: upvote_album_path(@album.id),         
                                   edit_name: @album.name, 
                                   edit_name_path: edit_album_path(@album.id), 
                                   name_path: album_path(@album.id), 
                                   view_name: "Albums", 
                                   all_names_path: albums_path} %>
```

app/views/albums/index.html.erb
```diff
@@ -0,0 +1,11 @@
+<table class="table table-hover">
```
kariabancroft, line 1: I would recommend including this whole table structure within the index_content shared partial and utilize the list of each model type as a local, rather than duplicating the table structure within each index view

app/views/shared/_new_form.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= form_for subject, html: {class: "form"}, url: path do |f| %>
```
kariabancroft: Since this form is very close to the edit for, I'd recommend trying to create _one_ shared form view that can handle both edit and new

config/routes.rb
```diff
@@ -0,0 +1,62 @@
...
+  # You can have the root of your site routed with "root"
+  root 'users#index'
+
+  resources :movies, :books, :albums
+
+  patch 'movies/:id/upvote' => "movies#upvote", as: "upvote_movie"
+  patch 'books/:id/upvote'  => "books#upvote",  as: "upvote_book"
+  patch 'albums/:id/upvote' => "albums#upvote", as: "upvote_album"
+
+  # Example of regular route:
```
kariabancroft, line 14: Once you feel comfortable with your routes, you should delete the default routes that are included in these comments

coverage/.resultset.json
```diff
@@ -0,0 +1,1312 @@
+{
+  "RSpec": {
+    "coverage": {
```
kariabancroft, line 3: All coverage information should be included in your gitignore file

spec/controllers/albums_controller_spec.rb
```diff
@@ -0,0 +1,224 @@
+require 'rails_helper'
+
+RSpec.describe AlbumsController, type: :controller do
+
+# INDEX ACTION__________________________________________________________________
+
+  describe "GET #index" do
```
kariabancroft, line 7: All of these index-based tests are totally independent of the controller, so this would be a great start in attempting to implement a shared example for all controllers.

spec/controllers/albums_controller_spec.rb
```diff
@@ -0,0 +1,224 @@
...
+      expect(response).to render_template("index")
+    end
+  end
+
+# SHOW ACTION___________________________________________________________________
+
+  describe "GET #show" do
+    before :each do
+      @album = Album.create(album_params[:album])
+      @album.ranking = 0
```
kariabancroft, line 25: This would be a great thing to put as a default value within your model, then `ranking` would always be set on every create even if you didn't explicitly set it on the model

spec/controllers/albums_controller_spec.rb
```diff
@@ -0,0 +1,224 @@
...
+  end
+
+# UPDATE ACTION_________________________________________________________________
+
+  describe "PATCH #update" do
+    before :each do
+      @album = Album.create(album_params[:album])
+    end
+
+    let(:album_params) do
```
kariabancroft, line 143: It looks like this would be a great place to attempt to refactor to use factorygirl in the future, since each time i've seen these album params, which create a model object, the params are the same (or very very similar).

spec/support/shared_examples_for_models.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+
+RSpec.shared_examples "a collection" do
```
kariabancroft, line 3: You did a nice job with this shared example for your models.

kariabancroft:
You did a really nice job on this. There are a few things that I would attempt to clean up in the controllers, but overall they look good. I think the # of the locals you use with your partials is okay, I'd just separate it out into separate lines to abide by the style guide and therefore make it easier to read.

app/controllers/albums_controller.rb
vikshab: moved this logic to the model - `scope :desc_order` (for all models )

app/controllers/albums_controller.rb
vikshab: Moved `@album = Album.find(params[:id])` into private method `find_album` and created `before-action :find_album, only: [:show, :edit, :update, :destroy :upvote]`` Did the same in other controllers

app/controllers/albums_controller.rb
vikshab: Changed `create_params` to `album_params`. Did the same changes in other controllers

app/controllers/albums_controller.rb
vikshab: I used `before_action`

app/controllers/books_controller.rb
vikshab, line 1: Fixed

app/controllers/movies_controller.rb
vikshab, line 1: Fixed

app/models/album.rb
vikshab: I changed the `scope :best, -> { order('ranking DESC').limit(10) }` to `scope :best, -> { order('ranking DESC') }` for all the models, created a constant `LIMIT` in `users_controller`, so 
the index method looks like `@movie = Movie.desc_order.limit(LIMIT)`

app/views/albums/show.html.erb
vikshab: Fixed

app/views/shared/_new_form.html.erb
vikshab: Created just one partial `_form.html.erb` for new and edit

config/routes.rb
vikshab, line 14: Did you mean all the commented out routes examples?

coverage/.resultset.json
vikshab, line 3: Added to .gitignore `coverage/`

kariabancroft:
Nice job refactoring your controllers and scopes!

PR #15, opened 2015-07-11
Media Ranker - Corinne Pingul

corinnepingul:
I have mixed feelings about this project. I really liked adding bootstrap and how much easier it made styling! Woohoo. But, I think that I could have dried up my code a bit better. I just learned about polymorphic paths and think that could have solved my if statement repetition problems. If I had more time, I would go back to try and implement those. I have also only written specs for my movie model and movies controller. I want to try to figure out shared examples for specs so that I don't have to re-write a ton of specs. I hope to add that to this pull request.


.gitignore
```diff
@@ -1,35 +1,19 @@
...
+/db/*.sqlite3
+/db/*.sqlite3-journal
 
-# if using bower-rails ignore default bower_components path bower.json files
-/vendor/assets/bower_components
-*.bowerrc
-bower.json
+# Ignore all logfiles and tempfiles.
+/log/*
+!/log/.keep
```
kariabancroft, line 16: I'd recommend adding all .keep files to your gitignore

.gitignore
```diff
@@ -1,35 +1,19 @@
...
+/tmp
 
-# Ignore pow environment settings
-.powenv
+coverage/
```
kariabancroft, line 19: Nice!

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    @album = Album.find(params[:id])
+  end
+
+  def edit
+    @album = Album.find(params[:id])
+    @action = "Edit"
+  end
+
+  def update
+    @album = Album.find(params[:id])
```
kariabancroft, line 17: You use this one line in a lot of places to retrieve an individual album item. I'd recommend creating a private method that does this and then apply it to a before_action

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    if params[:album].nil? == true
+      @album.rank += 1
+      @album.save
+      render :show
+    else
+      @album.update(album_params)
+      if @album.save
```
kariabancroft, line 25: The `update` method does the save operation for you so calling the save method again is redundant

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    redirect_to albums_path
+  end
+
+  def new
+    @album = Album.new
+    @action = "New"
+  end
+
+  def create
+    params[:album][:rank] = 0
```
kariabancroft, line 47: It would be worthwhile to set a default value for the rank in the model so that you don't need to explicitly set it anywhere else

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def edit
+    @album = Album.find(params[:id])
+    @action = "Edit"
+  end
+
+  def update
+    @album = Album.find(params[:id])
+
+    if params[:album].nil? == true
```
kariabancroft, line 19: Using the `.nil?` method call with automatically return a T/F so you don't really need the `== true` in addition to that

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,64 @@
+class BooksController < ApplicationController
```
kariabancroft, line 1: Same comments from the Album controller apply to this one

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,67 @@
+class MoviesController < ApplicationController
```
kariabancroft, line 1: Same comments from the Album controller apply to this one

app/models/album.rb
```diff
@@ -0,0 +1,8 @@
+class Album < ActiveRecord::Base
+  # Validations --------------------------------
+  validates :title, presence: true
+
+  # Scope --------------------------------------
+  scope :rank_order, -> { order(rank: :desc)}
+  scope :rank_order_top_ten, -> { order(rank: :desc).limit(10)}
```
kariabancroft, line 7: I really like that you're using a scope for this, but I would rename it to `rank_order_top` and exclude the `ten` so if you ever changed the # of values included in your top ranking you wouldn't have to change the name of your scope.

kariabancroft, line 7: I think you can also utilize the `rank_order` scope within this one rather than duplicating the logic

app/views/shared/_edit_media.html.erb
```diff
@@ -0,0 +1,18 @@
+<h2>
+  <%= @action %> <%= medium.class %>
```
kariabancroft, line 2: I might rename this shared partial since it is used for creation and editing purposes

app/views/books/show.html.erb
```diff
@@ -0,0 +1,11 @@
+<div class="col-lg-12">
+  <%= render partial: "shared/show", locals: { medium: @book,
+                                                      creator: @book.author,
+                                                      created_by: "Written by: " }%>
+
+  <%= button_to "Upvote", book_path, method: :put, class: "btn btn-primary", style: "margin-bottom: 10px" %>
+  <%= link_to "Edit #{@book.title}", edit_book_path, class: "btn btn-default" %>
+  <%= link_to "Delete", book_path, method: :delete, data: { confirm: "Are you sure?"}, class: "btn btn-danger" %>
+  <%= link_to "View all Books", books_path, class: "btn btn-default" %>
```
kariabancroft, line 9: I think that implementing polymorphic paths would enable you to include these buttons/links in the shared show partial

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>C3ProjectsMediaranker</title>
```
kariabancroft, line 4: I'd try to get in the habit of renaming the default app title that rails generates for you

app/views/shared/_home_display.html.erb
```diff
@@ -0,0 +1,35 @@
+<div class="row">
+  <% @media.each do |all_of_one_media| %>
```
kariabancroft, line 2: I like the way you utilized the `@media` array to contain each type of media but it doesn't really seem like it saved you all that much duplication. Do you think you could still do this with just one if statement?

app/views/shared/_show.html.erb
```diff
@@ -0,0 +1,11 @@
+<h1>
+  <%= medium.title %>
+  <small>
+    <%= created_by %>
+    <% unless creator.nil? == true %>
```
kariabancroft, line 5: Same comment about `.nil?` returning a T/F

app/views/welcome/index.html.erb
```diff
@@ -0,0 +1,4 @@
+<% @media_types = ["Movies", "Books", "Albums"] %>
```
kariabancroft, line 1: I think it would make sense for this to be in the controller rather than in this view

config/routes.rb
```diff
@@ -0,0 +1,59 @@
...
+  # Example of named route that can be invoked with purchase_url(id: product.id)
+  #   get 'products/:id/purchase' => 'catalog#purchase', as: :purchase
+
+  # Example resource route (maps HTTP verbs to controller actions automatically):
+  #   resources :products
+  resources :movies
+  resources :books
+  resources :albums
+
+  # Example resource route with options:
```
kariabancroft, line 20: Once you feel comfortable with your routes I'd remove the default comments

spec/controllers/albums_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe AlbumsController, type: :controller do
+  it_behaves_like 'a MediaController', Album, :album, :albums_path, :album_path
```
kariabancroft, line 4: Another way to pass in variable is to include a `do` block and then include these variables in `let` statements

spec/support/controllers/shared_example.rb
```diff
@@ -0,0 +1,115 @@
...
+    end
+  end
+
+  describe "PUT #update" do
+    before :each do
+      @medium = model_type.create(title: "HPCOS", description: "blah", rank: 0)
+    end
+
+    let(:description_params) { { id: @medium.id, model_symbol => { description: "yes" } } }
+    let(:title_params) { { id: @medium.id, model_symbol => { title: "" } } }
```
kariabancroft, line 17: Nice job creating these params dynamically

spec/support/controllers/shared_example.rb
```diff
@@ -0,0 +1,115 @@
...
+
+    context "When upvote is pressed" do
```
kariabancroft, line 19: This context is describing the physical action but the test is just testing the request itself, so you'd want to rename this to refer to the thing that is happening rather than the physical action

spec/support/controllers/shared_example.rb
```diff
@@ -0,0 +1,115 @@
...
+      # positive test - when upvote is pressed, it only updates the rank
+      it "updates the rank by one when upvote is pressed" do
+        put :update, id: @medium.id
+
+        expect(model_type.find(@medium.id).rank).to eq 1
+      end
+    end
+
+    context "Editing the form" do
```
kariabancroft, line 28: Same comment as above - this test is working on the edit action, the form is not involved at this point in the process

spec/support/controllers/shared_example.rb
```diff
@@ -0,0 +1,115 @@
...
+      end
+
+      # when it updates the form information, the upvote isn't changed
+      it "doesn't change the rank when you update the description" do
+        put :update, description_params
+
+        expect(model_type.find(@medium.id).rank).to eq 0
+      end
+
+      context "The record has a name!" do
```
kariabancroft, line 43: The name associated with this context is a bit confusing based on what is actually testing

kariabancroft:
You did a really nice job with this - only a few things you could clean up in your specs and controllers

PR #18, opened 2015-07-11
Media Ranker

catchingash:
This completes MediaRanker.

I started with a 3-model, 3-controller design, but wanted to explore something different than what I've done before, so I refactored into a single model, single controller (+ a controller for the landing page) design. This made for some unsightly switch cases that I put in the helper and in the model -- if there's a better way to handle this, please let me know! Polymorphic paths don't work with a single model, and I didn't find other solutions when I went looking.

I'm pretty happy with my test coverage, but I am missing a lot of negative testing.


app/controllers/welcome_controller.rb
```diff
@@ -0,0 +1,7 @@
+class WelcomeController < ApplicationController
+  def index
+    @movies = Medium.best_movies(5)
+    @books = Medium.best_books(5)
+    @albums = Medium.best_albums(5)
```
urmilan, line 5: It would be nice to factor the `5` out into a configuration file, or at least into a constant.

app/models/medium.rb
```diff
@@ -0,0 +1,43 @@
+class Medium < ActiveRecord::Base
+  scope :movies, -> { where(format: "movie") }
+  scope :books, -> { where(format: "book") }
+  scope :albums, -> { where(format: "album") }
+
+  scope :best, -> (total) { order('votes DESC').limit(total) }
+  scope :best_movies, -> (total) { movies.order('votes DESC').limit(total) }
+  scope :best_books, -> (total) { books.order('votes DESC').limit(total) }
+  scope :best_albums, -> (total) { albums.order('votes DESC').limit(total) }
```
urmilan, line 9: :+1: I like the scopes built on top of scopes!

app/models/medium.rb
```diff
@@ -0,0 +1,43 @@
+class Medium < ActiveRecord::Base
+  scope :movies, -> { where(format: "movie") }
+  scope :books, -> { where(format: "book") }
+  scope :albums, -> { where(format: "album") }
+
+  scope :best, -> (total) { order('votes DESC').limit(total) }
+  scope :best_movies, -> (total) { movies.order('votes DESC').limit(total) }
```
urmilan, line 7: Instead of repeating `.order('votes DESC')` here, could you reuse the `best` scopes you already created and apply it to `movies`?

app/models/medium.rb
```diff
@@ -0,0 +1,43 @@
...
+  scope :best_books, -> (total) { books.order('votes DESC').limit(total) }
+  scope :best_albums, -> (total) { albums.order('votes DESC').limit(total) }
+
+  validates :title, presence: true
+  validates :format, presence: true, inclusion: { in: ["movie", "book", "album"], message: "\"book\", \"movie\", and \"album\" are the only allowed formats." }
```
urmilan, line 12: In general, the strings `"movie"`, `"book"` and `"album"` are repeated a bunch in the code. It would probably be good to declare these as constants (or maybe even enums?) to make sure you don't tripped up by an errant typo somewhere...

app/models/medium.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def self.upvote(object)
```
urmilan, line 14: What was the reasoning behind making this a class method?

urmilan, line 14: When you have a method that mutates an object, it's good practice in Ruby to end that method name with a bang eg. `def self.upvote!`.

app/models/medium.rb
```diff
@@ -0,0 +1,43 @@
...
+    format == "movie"
+  end
+
+  def book?
+    format == "book"
+  end
+
+  def album?
+    format == "album"
+  end
```
urmilan, line 29: I possibly just missed something, but I didn't actually see these methods used anywhere except for the tests?

app/helpers/media_helper.rb
```diff
@@ -0,0 +1,45 @@
+module MediaHelper
+  def object_path(obj)
+    case obj.format
+    when "movie"
+      movie_path(obj)
+    when "book"
+      book_path(obj)
+    when "album"
+      album_path(obj)
+    end
```
urmilan, line 10: Your `case` statements should have an `else` to handle the error case in some smart way (eg. throwing an error). Otherwise (I think) if you pass weird values to these methods they will no-op silently, which can be really hard to debug in a large working application.

urmilan:
I really loved that you experimented with the single model design in this project! So cool! The switch statements are a bit clunky, as you acknowledged, but it saved you from writing a lot of boilerplate code (which makes enhancement/modification much more efficient in the future).

catchingash:
Thanks, @urmilan !! All great points :)

app/helpers/media_helper.rb
catchingash, line 10: @urmilan, I completely agree. I wasn't sure what kind of error was appropriate for a view helper, though -- should this raise an exception?

PR #20, opened 2015-07-11
Media Ranker

QuartzBrandi:
Here is my Media Ranker project.

I personally didn't feel like I did anything special with this project. Things I would have liked to have done: implemented polymorphic paths & created shared rspecs for the controllers. I also wanted to figure out how to do custom bootstrap, but I wasn't sure how and didn't have time to even look into it.


Gemfile
```diff
@@ -0,0 +1,52 @@
...
+  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
+  gem 'byebug'
+
+  # Access an IRB console on exception pages or by using <%= console %> in views
+  gem 'web-console', '~> 2.0'
+
+  # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
+  gem 'spring'
+
+  # Rspec tests--IN RAILS!
```
jnf, line 48: lolololol

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+    votes = @album.rank
+    votes += 1
+    @album.update(rank: votes)
+
+    super
+  end
+
+  private
+
+  def permit_params
```
jnf, line 55: I know that we haven't really gotten into using a buncha these in a controller, but, in larger projects it is pretty common to have several/many of these kinds of methods. So you could end up with `create_album_params` and `update_album_details_params` and `update_album_owner_params` (or whatevs).

So, moving forward, think about what these params are for, and name the method to reflect that intended use.

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+    Album.find(params[:id]).destroy
+
+    super
+  end
+
+  def upvote
+    @album = Album.find(params[:id])
+    votes = @album.rank
+    votes += 1
+    @album.update(rank: votes)
```
jnf, line 48: When we did Woot'n'Grumble (attorneys at law), we saw the `increment!` ActiveRecord method, so I'm not gonna bring it up here. I didn't know about it either, so my approach was a lot like yours.

I think you could simplify this a little bit though by using the `+=` operator. Something like...

``` ruby
@album = Album.find(params[:id])
@album.rank += 1
@album.save
```

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def new
+    @album = Album.new
+  end
+
+  def create
+    @album = Album.new(permit_params)
+    @album.rank = 0
+    if @album.save
+      super
```
jnf, line 19: This is a really interesting use of `super`. It's definitely not common practice, but I think it's pretty clever.

You're recognizing common patterns in the application, and this could defs be a stepping stone towards a refactor.

app/models/album.rb
```diff
@@ -0,0 +1,11 @@
+class Album < ActiveRecord::Base
+
+  validates :name, presence: true
+  # validates :artist
+  # validates :description
+  validates :rank, presence: true, numericality: { only_integer: true }
+
+  scope :best_first, -> (total) { all.order('rank DESC').limit(total) }
+  scope :best_first_all, -> { all.order('rank DESC') }
```
jnf, line 9: The `all` calls in these scopes are technically unnecessary.`{ order('rank DESC') }` works just as well, but is definitely not obvious.

app/views/albums/index.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="row">
+  <div class="col-lg-12">
+    <table class="table table-hover">
+      <tbody>
+
+        <%= render partial: "shared/list_headers" %>
+
+        <%= render partial: "shared/list", collection: @albums, as: :object,
```
jnf, line 8: Awesome use of `collection`. One of my favorite hacks in Rails is rendering over a collection. :+1:

app/views/shared/_show.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    <%= by + object_person %>
+  </small>
+</h1>
+
+<h4>Ranked: <%= object.rank %></h4>
+<p><%= object.description %></p>
+
+<%= button_to "Upvote", upvote_route, class: "btn btn-primary", method: "patch" %>
+<div></div>
+<%= link_to "Edit #{object.name}", edit_single_route, class: "btn btn-default", method: "get" %>
```
jnf, line 13: You mentioned this in your PR description, but `polymorphic_path` would simplify a lot of this for you. I mention it here so, should you ever want to refactor, this would be a good place to start.

spec/controllers/albums_controller_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+      end
+
+      it "updates a album" do
+        patch :update, :id => @original_album.id, album: {
+          name: "edited test title",
+          artist: "test artist",
+          description: "test description"
+        }
+        # @original_album.reload
+        expect(Album.find(1).name).to eq "edited test title"
```
jnf, line 116: This expectation has a dependency on the album id. There's no guarantee that the id would be 1. It's an assumption that the attached data store is relational, uses integer keys, and has no other records. That'll be true _most_ of the time, but, to make it safes, you could rewrite the expectation to something like `expect(Album.find(@original_album.id).name).to`, which does the same thing, but doesn't make the assumptions about the album id.

spec/controllers/books_controller_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+      end
+
+      it "creates a book" do
+        post :create, book_params
+        expect(Book.count).to eq 1
+      end
+
+      it "redirects to the book show page" do
+        post :create, book_params
+        expect(subject).to redirect_to(books_path)
```
jnf, line 56: Hmm. I would expect the _book show page_ to be `book_path(book.id)`. Looking at the controller, I see that it is redirecting to `books_path`, so just the test description is inaccurate.

spec/support/media_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+      expect(described_class.best_first(10).first).to eq(described_class.find_by(rank: 101))
+    end
+
+    it "includes the highest ranked movies (designated by total argument)" do
+      expect(described_class.best_first(10).count).to eq 10
+      expect(described_class.best_first(10)).to include(@included_media)
+    end
+
+    it "excludes the lowest ranked movies (designated by total argument)" do
+      expect(described_class.best_first(10)).to_not include(@excluded_media)
```
jnf, line 58: Since this is the only place you're using `@excluded_media`, I'd consider moving it's construction to this test, rather than recreating it in every test in the `before` block. Ditto the `@included_media` above.

spec/support/media_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    before :each do
+      ranks = [2, 1, 3, 3, 4, 5, 0, 100, 50, 101, 24, 35, 0, 0, 40, -1, 51, -20]
+      ranks.each do |rank|
+        described_class.create(name: "Test #{rank}", rank: rank)
+      end
+    end
+
+    it "ranks by highest number first" do
+      expect(described_class.best_first_all.first).to eq(described_class.find_by(rank: 101))
+    end
```
jnf, line 74: Another good test here would be verifying that the scope included all the records by having an expectation like...

``` ruby
expect(described_class.best_first_all.count).to eq described_class.count
```

jnf:
Hey hey! Good job on this project. You may not think there's "anything special" here, but I certainly do. Your code is well organized, clear in its intent, and has a pretty good set of foundational tests. I've left a few notes for you to review, but, overall, I'm very happy with your code. :+1:

app/controllers/albums_controller.rb
QuartzBrandi, line 55: Good to know.

spec/controllers/books_controller_spec.rb
QuartzBrandi, line 56: Oops, that's what I get for copying and pasting. This is supposed to be describing the create action.

PR #21, opened 2015-07-11
Media Ranker project

a-lmx:
Here is my completed media ranker project. I spent time thinking about the design of the project, and settled on using three different classes for each different type of media to be object oriented, and so that I could use Rails magic in a straight-forward way. I tried to figure out how to implement an abstract class that could house the identical model logic and have the concrete classes inherit from that (without doing single table table inheritance), but I couldn't get it to work. That's on my list of things to figure out in Rails going forward.

I tried to be comprehensive in my tests, and would love feedback on what else I should test, or which tests are overkill. I pulled the model and controller tests out into shared examples, and was happy to be able to pass in only one variable from which I could access all the necessary paths & models, etc., but think the way I did it is pretty hack-y.


GreyKn:
Good thinking about abstracting the models. You could (potentially) have created an abstract model class like this:

models/basemodel.rb

```
class BaseModel < ActiveRecord::Base
  # Validations ----------------------------------------------------------------
  validates :title, presence: true
  validates :rank, numericality: { only_integer: true, greater_than_or_equal_to: 0 }

  # Scopes ---------------------------------------------------------------------
  TOP_NUMBER = 10

  scope :ranked,      -> { order(rank: :desc) }
  scope :top_ranked,  -> { ranked.limit(TOP_NUMBER) }

  def add_vote
    update(rank: self.rank + 1)
  end
end
```

models/album.rb

```
class Album < BaseModel
  set_table_name 'Album'
end
```

models/book.rb

```
class Book < BaseModel
  set_table_name 'Book'
end
```

models/movie.rb

```
class Movie < BaseModel
  set_table_name 'Movie'
end
```

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,51 @@
+class BooksController < ApplicationController
+  before_action :set_book, only: [:show, :edit, :update, :upvote, :destroy]
```
GreyKn, line 2: Good use of the before_action to reduce redundant code for getting the book model. :+1: 

You're setting the before_action using `only` on 5 out of 7 actions. You could say save some type by using `except` and list only 2.

spec/controllers/albums_controller_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+require 'support/shared_examples_for_media_controllers'
+
+RSpec.describe AlbumsController, type: :controller do
+  let(:model) { Album }
```
GreyKn, line 5: Good use of using a `let` to feed a variable into shared_examples. :+1: 

One thing you could do is define a model class method in your controller, and define the model there for use with tests. So you could do:

```
  class AlbumsController < ActiveController::Base
     def self.model
       Albums
     end
  end
```

This way, the controller controls what model is being used instead of your test, by doing `described_class.model`

spec/support/shared_examples_for_media_controllers.rb
```diff
@@ -0,0 +1,197 @@
+require 'spec_helper'
+
+RSpec.shared_examples "a MediumController" do
+  let(:model_string) { model.to_s.downcase }
+  let(:medium_path) { model_string + "_path" }
+  let(:mediums_path) { model_string + "s_path" }
+  let(:medium_symbol) { model_string.to_sym }
+
+  describe "GET #index" do
+    before :each do
```
GreyKn, line 10: Great use of `before`

spec/support/shared_examples_for_media_controllers.rb
```diff
@@ -0,0 +1,197 @@
...
+      expect(response).to have_http_status(200)
+    end
+
+    it "renders index template" do
+      expect(response).to render_template("index")
+    end
+  end
+
+  describe "GET #show" do
+    let(:medium) { model.create(title: 'some title', creator: 'some creator') }
```
GreyKn, line 25: Not sure how the test stuff is working here, but as an FYI, it's good practice, whenever you create a DB entry for your test, to delete the DB entry after you're done with it. It helps prevents side effects in other tests.

You could use an after block:

```
after(:each) do
  medium.destroy
end
```

spec/support/shared_examples_for_media_controllers.rb
```diff
@@ -0,0 +1,197 @@
...
+    end
+
+    it "creates a new medium" do
+      expect(assigns(medium_symbol)).to be_a_new(model)
+    end
+  end
+
+  describe "POST #create" do
+    # positive test - medium params are valid
+    context "valid medium params" do
```
GreyKn, line 61: Cool, you're using `context` for what it's good for

spec/support/shared_examples_for_media_models.rb
```diff
@@ -0,0 +1,112 @@
...
+  let(:medium) { described_class.new }
+
+  describe "model validations" do
+    it "requires a title, all the time" do
+      expect(medium).to_not be_valid
+      expect(medium.errors.keys).to include(:title)
+    end
+
+    context "validating rank" do
+      ["pork", 3.14, -1].each do |invalid_rank|
```
GreyKn, line 13: Good use of loops to write 3 tests in one.

app/models/album.rb
```diff
@@ -0,0 +1,15 @@
+class Album < ActiveRecord::Base
+  # Validations ----------------------------------------------------------------
+  validates :title, presence: true
+  validates :rank, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+
+  # Scopes ---------------------------------------------------------------------
+  TOP_NUMBER = 10
```
GreyKn, line 7: Good use of a constant to hold a static referenced value

app/views/albums/edit.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= render partial: 'shared/form',  locals: { medium: @album,
```
GreyKn, line 1: I noticed you're repeating your views for all your types. It's possible to have used a single shared set of views if you got a little creative with naming your instance variables.

GreyKn:
Hey Alice,

Really good job on this project. Super clean formatting and your tests are great.

The only things I would add is that you could have generic-ized your controllers and your views and written a little bit less code.

But overall, rockin job on this :100: 

Let me know if you have any questions.
-Chef

app/controllers/books_controller.rb
a-lmx, line 2: Makes sense. I'll add this to my list of parts of the code to optimize at the end of a project.

spec/controllers/albums_controller_spec.rb
a-lmx, line 5: I like this!

spec/support/shared_examples_for_media_controllers.rb
a-lmx, line 25: I thought objects created in tests were deleted at the end of each test. And that before :each blocks were executed within the tests. I'll look into this further - thanks.

a-lmx:
Thanks for all the comments, Chef. I look forward to trying out an abstract class in a future project. I'll merge this PR and submit a new one using your suggestions for some refactoring.

PR #22, opened 2015-07-11
MediaRanker complete

carjug:
hi!
This is my MediaRanker project. I completed all necessary requirements. A thing I'm really proud of is that I extracted all my view files into partials. I also extracted my model tests into a shared_example controller spec file. Neat!

I didn't do a ton of testing, as that wasn't my main focus. My interests were more in DRYing up my code and making sure everything ran properly. I am interested in testing so long as it drives the development of my code, but with this assignment I found it hard to care about testing retroactively.

Thanks for you feedback!!!! It is invaluable!


.gitignore
```diff
@@ -1,35 +1,17 @@
...
-bower.json
+# Ignore the default SQLite database.
+/db/*.sqlite3
+/db/*.sqlite3-journal
 
-# Ignore pow environment settings
-.powenv
+# Ignore all logfiles and tempfiles.
+/log/*
+!/log/.keep
```
kariabancroft, line 16: I'd recommend adding all .keep files to your gitignore

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  def show
+    @album = Album.find(params[:id])
+  end
+
+  def edit
+    @album = Album.find(params[:id])
+  end
+
+  def update
+    @album = Album.find(params[:id])
```
kariabancroft: It looks like you are reusing this line in many places in your controller. This would be a great place to use a `before_action` and write a private method that would `find` each album.

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def upvote
+    @album = Album.find(params[:id])
+
+    @album.rank += 1
+
+    @album.save
+
+    redirect_to(:back) rescue redirect_to album_path
```
kariabancroft, line 40: This is cool!

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def destroy
+    @album = Album.destroy(params[:id])
```
kariabancroft: Do you need to set the result of this `destroy` call to an instance variable?

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    redirect_to albums_path
+  end
+
+  private
+
+  def create_params
```
kariabancroft: I'd recommend re-naming this method since it is used for more than just the creation operation

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,56 @@
+class AlbumsController < ApplicationController
+
+  def index
+    @albums = Album.all
+
+    @albums = @albums.order('rank DESC')
```
kariabancroft: I think it would be better to include the ordering of the results in the model rather than here

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,58 @@
+class BooksController < ApplicationController
```
kariabancroft, line 1: Same comments from the AlbumsController

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,59 @@
+class MoviesController < ApplicationController
```
kariabancroft, line 1: Same comments from the AlbumsController

app/controllers/user_controller.rb
```diff
@@ -0,0 +1,13 @@
+class UserController < ApplicationController
+
+  def index
+    @movies = Movie.best
```
kariabancroft, line 4: I like the way that you used a scope for this. It would be a bit more extendable if you had it accept the parameter for the # of records you wanted to include. Then you could use a single constant to store the # of records you want and if the top # ever changed, you would only have to change it in one place rather than 3.

spec/support/shared_examples_for_models.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  context "validating rank" do
+    it "requires rank to be an integer" do
+      object = described_class.new(name: "Whatev", rank: "hi")
+
+      expect(object).to_not be_valid
+      expect(object.errors.keys).to include(:rank)
+    end
+
+    it "requires rank to be less than 200" do
```
kariabancroft, line 30: I really like the fact that you included this in your model definition and also wrote a test for this. Nice job.

app/views/shared/_new.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    </div>
+
+
+    <div class="form-group">
+      <%= f.label :rank, "Rank: " %>
+      <%= f.number_field :rank, class: "form-control" %>
+    </div>
+  </fieldset>
+
+  <%= f.submit "Add #{object.class}", class: "btn btn-default" %>
```
kariabancroft, line 28: This partial looks the same as the Edit partial except for this one line. I think this would be a great place to combine both into a `_form` partial and then reduce the views even more

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>C3ProjectsMediaranker</title>
```
kariabancroft, line 4: Try to get in the habit of updating the default title of your app whenever you work on a new project

app/views/movies/index.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= render partial: "shared/index.html.erb", locals: { path: "movies", collection: @movies } %>
+
+<%= link_to "View All Media", root_path, class: "btn btn-default", role: "button" %>
```
kariabancroft: This line would be a great place to implement a view helper since it seems like this exact "View All Media" section is repeated in multiple places

app/views/shared/_index.html.erb
```diff
@@ -0,0 +1,14 @@
+<table class="table table-condensed table-hover">
+  <th>Ranking</th>
+  <th>Name</th>
+  <th>Upvote</th>
+  <tbody>
+    <% collection.each do |item| %>
+    <tr>
+      <td><%= "Ranked as: #{item.rank}" %></td>
+      <td><%= link_to "#{item.name}", "#{path}/#{item.id}" %></td>
```
kariabancroft, line 9: I know you used polymorphic paths in some other places, I'm wondering if you could've used that here as well

app/views/shared/_main.html.erb
```diff
@@ -0,0 +1,13 @@
...
+ <h3><%= "Top #{object.name}s" %></h3>
+
+  <% object.each do |detail| %>
+    <h2>
+      <%= link_to "#{detail.name} ", polymorphic_path(detail) %>
+      <small><%= "Ranked as: #{detail.rank}" %></small>
+    </h2>
+  <% end %>
+
+<%= link_to "View All #{object.name}s", polymorphic_path("#{object.name.downcase}s"), class: "btn btn-default", role: "button" %>
```
kariabancroft, line 11: You might be able to use `pluralize` rather than tacking on the `s` manually

config/routes.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  patch '/albums/:id/upvote' => 'albums#upvote', as: 'album_upvote'
+
+  patch '/books/:id/upvote' => 'books#upvote', as: 'book_upvote'
+
+  patch '/movies/:id/upvote' => 'movies#upvote', as: 'movie_upvote'
+
+
+
+  # Example resource route with options:
```
kariabancroft, line 44: Once you feel good with your routes you should delete any of these routes that are included in the comments as defaults

db/schema.rb
```diff
@@ -0,0 +1,48 @@
...
+# system, you should be using db:schema:load, not running all the migrations
+# from scratch. The latter is a flawed and unsustainable approach (the more migrations
+# you'll amass, the slower it'll run and the greater likelihood for issues).
+#
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema.define(version: 20150706234337) do
+
+  create_table "albums", force: :cascade do |t|
+    t.string   "name"
```
kariabancroft, line 17: These fields should probably have some additional database requirements (like `null: false`)

spec/controllers/albums_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      album.update(name: "Tragic Kingdom")
+
+      expect(album.name).to eq("Tragic Kingdom")
+    end
+  end
+
+  describe "PATCH #upvote" do
+    it "increases the album rank by 1" do
+      album = Album.create(name: "Tragic Kingdom", rank: 99)
```
kariabancroft, line 39: I think that it might make sense to put this Album creation in a `let` statement since it doesn't seem like you necessarily need to have different album instances for each of these tests

spec/support/shared_examples_for_models.rb
```diff
@@ -0,0 +1,38 @@
+require 'rails_helper'
+
+RSpec.shared_examples "an object" do
```
kariabancroft, line 3: I like that you made a shared example here, but I would use a name other than "object" since it's so generic across... everything

kariabancroft:
I really like the way you used polymorphic paths as well as your shared partials. There are still a few things you can clean up in the controller and tests, but nothing major. Next opportunity for refactoring on this project would be to use a shared example for your controller tests. Nice job overall!

carjug:
Hiiii. I just realized I never merged my previous PR. 
Ahhhh. I'm going to hold off on merging just in case it's a big deal.
Sorry about that! The last 12 commits are for the refactoring assignment.

kariabancroft:
You did a really nice job refactoring in your controllers!

PR #23, opened 2015-07-11
mar - Media ranker

camarleigh:
well, wow.

So I had a light bulb moment with this project: I gotta stop comparing my code to other people's and be my own measuring stick. Translation: it's been 9 weeks, and for someone who walked in the door with NO coding experience what-so-EVER, I feel like this project didn't go half bad. 

Details:

I tried to pay close attention to RESTful routing and CSS details with this project. While I changed the image in the banner, MOST of the rest of the project should mirror the media ranker heroku master model we were given, and my site is responsive. 

Yay bootstrap! Until I get the hang of Rails (including routing) I like the ease of Bootstrap and how it took a lot of design decisions I would have wasted hours on completely out of my hands. I'll get more selective with CSS with time, I'm sure, but for now, yay Bootstrap!

I'm feeling more confident with partials, forms, and blocks.  I attempted to DRY up lots of my code including the header, and the edit/create views. That being said, somewhere along the way, "edit/create album" stopped working - It's not saving to the database. It will respond to a route call, let you edit/or create, but when you press save, it simply redirects to the albums index page, with the new album or edits made are nowhere to be found. I looked through every part of Album MVC, but maybe my eyes are missing it - if you find it, lemme know. Edit/create book and movie are working fine....

I just learned about polymorphic partials and associations at 3:30 this afternoon. Man!, they are what I've been looking for. With every project we do, I feel like what I want to do  ( in this case combine models to one that identifies a bunch of "mediums") is juuuuust outta reach. I guess that's a good thing; it means I want more and feel I can improve. At least that's how I'm going to interpret it. (I've already created MVC to explore this avenue later - you'll find the media MVC empty though so ignore them).

RSPEC!!!!! - Okay I'm the weird chick who like, totally can't wait to explore testing in full. I'm looking forward to more exposure to it with future projects ( did I "say" that out loud?) Unfortunately, at the current submission for this project I only have three basic tests running for each MVC - 2 for each model (validation & scope) and one for each controller (recognizes :index). I will add to these in the near future, but all current tests are green! yay!

All in all, I enjoyed this project. I enjoy all our indiv. projects as they force me to evaluate how much I really know. I look forward to this review! Any and all suggestions are welcome, especially on testing, partials, and DRYing up code.

Thanks in advance,

m


app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+  def edit #GET one album to modify.
+   @album = Album.find(params[:id])   
+  end
+
+  def update #PATCH this updated album to the db
+    @album = Album.find(params[:id]) 
+    @album.name = permit_params[:album][:name]
+    @album.artist = permit_params[:album][:artist]
+    @album.description = permit_params[:album][:description]
+    @album.save
```
jnf, line 33: ActiveRecord's `update` method could save you some work here. What you're doing totally works well; `@album.update()` would just trim some lines.

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,57 @@
+class AlbumsController < ApplicationController
+
+  def index
+    @albums = Album.best(5)
+  end
+
+  def create
+    @album = Album.new(permit_params[:album])
+    @album.save
```
jnf, line 9: What happens if the album fails validation? It'd be good to introduce a fallback. Something like...

``` ruby
if @album.save
  redirect_to albums_path
else
  # what should happen here?
end
```

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    @album.artist = permit_params[:album][:artist]
+    @album.description = permit_params[:album][:description]
+    @album.save
+    
+    redirect_to(album_path(Album.find(@album.id))) 
+  end
+
+  def yes_vote # PATCH increase this number by one in the db
+   @album = Album.find(params[:id])
+   @album.increment!(:vote)
```
jnf, line 40: Nice!

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+  redirect_to(album_path(Album.find(@album.id)))
```
jnf, line 41: You've already got the album info in the `@album` instance variable, so you could simplify this line to just be `redirect_to album_path(@album)`.

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    @book = Book.new(permit_params[:book])
+    @book.save
+
+    redirect_to(books_path)
+  end
+
+  def new
+    @book= Book.new
+
+    render :new
```
jnf, line 17: This could be an implicit render. Given that you're using implicit renders in other actions (like `index` and `show`) I was momentarily thrown off by the explicit render here.

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def update #PATCH this updated book to the db
+    @book = Book.find(params[:id]) 
+    @book.name = permit_params[:book][:name]
+    @book.author = permit_params[:book][:author]
+    @book.description = permit_params[:book][:description]
+    @book.save
+    
+    redirect_to(book_path(Book.find(@book.id)))
```
jnf, line 35: My comments in the `AlbumController` apply here too; `update` is good savings here.

app/views/albums/index.html.erb
```diff
@@ -0,0 +1,23 @@
...
+        <th>Album Name</th>
+        <th>Upvote</th>
+      </tr>
+    </thead>
+    <tbody>
+      <% @albums.each do |album| %>
+      <tr>
+        <td>Ranked:<%= album.vote %></td>
+        <td><%= link_to "#{album.name}", album_path(album.id) %></td>
+        <td><%= button_to "Upvote", "/albums/#{album.id}/yes_vote", method: :patch, :class => "btn btn-default" %></td>
```
jnf, line 15: Constructing URLs with interpolation can be very brittle; always try to use the route prefixes/helpers. In this case, since you defined the `yes_vote` route as a member route in the routes file, you can use `yes_vote_path(album)` to generate the correct path for the button.

app/views/albums/show.html.erb
```diff
@@ -0,0 +1,21 @@
...
+
+    <p> <%= @album.description %> </p>
+
+    <br>
+        <%= button_to "Upvote", "#{@album.id}/yes_vote", method: :patch, :class => "btn btn-primary" %>
+    <div class="button_wrapper">
+        <%= link_to "Edit #{@album.name}", edit_album_path(@album.id), :class => " btn btn-default button_menu" %>
+        <%= link_to "Delete", album_path(@album.id), method: :delete, data: {confirm: "are you sure you want to delete this album?"}, :class =>  "btn btn-danger button_menu"%>
+        <%= link_to "View All Albums", albums_path, :class => "btn btn-default button_menu" %>
+        <%= link_to "View All Media", root_path, :class => "btn btn-default button_menu" %>
```
jnf, line 17: Good use of the route helpers/prefixes here. Swapping the Upvote path to `yes_vote_path(album)` would push this through the roof.

app/views/shared/_editcreate_album_form.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+    <%= f.label :description %>
+    <%= f.text_field :description %>
+
+
+  </fieldset>
+
+  <%= f.submit %>
+
+<% end %>
```
jnf, line 18: This comment applies too all three of the form partials. I love that you've isolated the form logic into reusable partials. Writing forms quickly gets complex, so managing that complexity in as few places as possible is a real boon to a project.

These forms are great, but I'd like to suggest one enhancement that would harden them against changes in the future. Right now, these forms rely on an instance variable to be in scope in order for the `form_for` to operate. If, in the future, the media type changed from @album to something else, this form would break.

You can generalize this form more by using a local variable instead of an instance variable. Here's an example:

``` erb
-# in the view that's calling the parital right now, you have
<%= render partial: 'shared/editcreate_movie_form' %>

-# here's a small change to pass in the variable the form requires
<%= render partial: 'shared/editcreate_movie_form', locals: { model: @album } %>
```

That change will make a local variable called `model` available for use in the partial. This passes the dependency on the instance variable up the chain of command from the partial to the view calling the partial. The partial will happily render a form for _any_ kind of object that responds to the fields `name`, `artist`, and `description`.

This would be an important first step to moving toward a singular model (`Medium`) and controller (`MediaController`) approach.

app/views/welcome/index.html.erb
```diff
@@ -0,0 +1,31 @@
...
+  <div class= "col-lg-12">
+  <div class="row">
+    <div class="col-md-4">
+
+      <h1>Top Movies</h1> 
+         <% @movies.each do |movie| %>
+          <h2> <%= link_to "#{movie.name}", movie_path(movie.id)  %> 
+          <small> Ranked: <%= movie.vote %> </small> </h2>
+        <%end %>
+        <p><%= link_to "See All Movies", movies_path, method: :get %><p>
```
jnf, line 11: `link_to` generates an anchor (`<a>`) html element. Anchors trigger `get` requests by default, so you only need to explicitly set the method when you want something other than `get`.

spec/models/album_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  describe "best scope" do
+    it "ranks all the albums by vote in descending order" do
+      album1= Album.create(name:'a', artist:'b', description: 'c', vote:30)
+      album2= Album.create(name:'a', artist:'b', description: 'c', vote:3)
+      album3= Album.create(name:'a', artist:'b', description: 'c', vote:20)
+
+      right_rank = [album1, album3, album2]
+
+      expect(Album.best(3)).to eq right_rank
```
jnf, line 22: This is a great test. Nice job. :+1:

jnf:
Overall you've done a really great job with this project. I've left some comments to give some thoughts on tweaks and enhancements. I'm not sure after browsing through here what's triggering the problem in the Albums workflow, but we can look at it together sometime this week. Chances are, we'll find the discrepancy by reading the logs.

Let me know if you've any questions about the project or my comments!

camarleigh:
Thanks so much, @jnf . I'm so glad this was assigned as homework cuz I've been wanting to get back to your review since I read it two weeks ago! But then bEtsy and  AdaCooks.... anyway, I need practic on all these things so thank you for pointing them oiyt. Gonna merge this and implement hose changes. in addition going to dry to implement some  before_action to help dry to some actions out.

Thanks again!

m

PR #24, opened 2015-07-12
Media Ranker

knguyen0710:
This is for the final submission of media ranker.

I had some trouble writing rspec tests for this project and being able to differentiate between which test should go where so that's where I ended up spending a lot of my time with this project. That, and I'm still not great at using github (probably because I get a little keyboard happy).

Other than that, I thought this was a great exploratory project to do on my own.


.gitignore
```diff
@@ -1,35 +1,17 @@
...
-bower.json
+# Ignore the default SQLite database.
+/db/*.sqlite3
+/db/*.sqlite3-journal
 
-# Ignore pow environment settings
-.powenv
+# Ignore all logfiles and tempfiles.
+/log/*
+!/log/.keep
```
kariabancroft, line 16: You should add all .keep files to your gitignore as well as the whole coverage folder

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def edit
+    @media = Album.find(params[:id])
+    @url = album_path
+    @method = :patch
+    @by = :artist
+  end
+
+  def update
+    @media = Album.find(params[:id])
```
kariabancroft, line 36: You're using this `find` method in a bunch of different methods here. This would be a great place to use a `before_action` to do this for you to clean this up.

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    @by = :artist
```
kariabancroft, line 37: This is another line that is repeated in multiple controller methods so this should also be in a before_action

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    @media.update(album_params)
+    if @media.save
```
kariabancroft, line 39: The `update` method automatically saves the update, so you don't need to call save again

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def destroy
+    Album.find(params[:id]).destroy
+    redirect_to albums_path
+  end
+
+  def vote
+    @media = Album.find(params[:id])
+    @media.ranking += 1
+    @media.save
```
kariabancroft, line 54: Does this `@media` object need to be an instance variable?

app/controllers/books_controller.rb
```diff
@@ -0,0 +1,59 @@
+class BooksController < ApplicationController
```
kariabancroft, line 1: Same comments from the AlbumsController

app/controllers/movies_controller.rb
```diff
@@ -0,0 +1,59 @@
+class MoviesController < ApplicationController
```
kariabancroft, line 1: Same comments from the AlbumsController

app/controllers/albums_controller.rb
```diff
@@ -0,0 +1,64 @@
+class AlbumsController < ApplicationController
+  def index
+    @albums = Album.all
+  end
+
+  def new
+    @media = Album.new
+    @url = albums_path
+    @method = :post
+    @by = :artist
```
kariabancroft, line 10: You are setting this instance variable but not passing it as a local from the albums/new.html.erb view. To me, this means that _none_ of these instance variables need to be passed as locals, since they are instance variables and will just go through without setting them explicitly as locals.

app/views/home/index.html.erb
```diff
@@ -0,0 +1,25 @@
+<div class="container-fluid">
+  <div class="row">
+    <div class="col-sm-4">
+      <h2>Bangin' Jams</h2>
+        <% @albums.limit(10).each do |album| %>
```
kariabancroft, line 5: It would make more sense to limit these records at the model level rather than here. Think about the case where they could be hundred of albums in the database, you wouldn't want that huge array to be passed from the controller to the view if you could just limit before it even got here to the view.

app/views/albums/index.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <tr>
+    <th>Ranking</th>
+    <th>Title</th>
+    <th>Vote</th>
+  </tr>
+  <% @albums.each do |album| %>
+  <tr>
+    <td><%= album.ranking %></td>
+    <td><%= link_to "#{album.title}", album_path(album.id), method: :get %></td>
+    <td><%= button_to "Vote", "/albums/upvote/#{album.id}", method: :get, class: "btn btn-default" %></td>
```
kariabancroft, line 12: You should use the named path rather than using the string version of the path

app/views/albums/new.html.erb
```diff
@@ -0,0 +1,10 @@
+<h1> New Album</h1>
+<ul>
+  <% @media.errors.each do |column, message| %>
```
kariabancroft, line 3: You're using the same instance variable `@media` in all of the new views, so I'm wondering why you decided not to include this error logic within the partial?

app/views/albums/show.html.erb
```diff
@@ -0,0 +1,10 @@
+<h1><%= @album.title %> <small>Artist:<%= @album.author %></small></h1>
+<h4>Ranked:<%= @album.ranking %></h4>
+<h5><%= @album.description %></h5>
+<div class="col-md-12">
+<%= button_to "Vote", "/albums/upvote/#{@album.id}", method: :get, class: "btn btn-primary" %>
```
kariabancroft, line 5: Here is another location where you should use the named path

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>C3ProjectsMediaranker</title>
```
kariabancroft, line 4: It is good to get in the habit of updating the default app title name that rails give you

app/views/shared/_form.html.erb
```diff
@@ -0,0 +1,21 @@
+<div class="col-sm-12">
+  <%= form_for @media, html: {class: "form"}, url: @url, method: @method do |f| %>
+  <div class="form-group">
```
kariabancroft, line 3: This `form-group` has nice indentation but the others in this view need some additional work

config/routes.rb
```diff
@@ -0,0 +1,67 @@
...
+  resources :albums
+
+  get 'books/upvote/:id' => 'books#vote'
+  resources :books
+
+  get 'movies/upvote/:id' => 'movies#vote'
+  resources :movies
+
+
+  # Example of regular route:
```
kariabancroft, line 19: Once you feel good with your routes, you should delete these default route comments that are included

db/schema.rb
```diff
@@ -0,0 +1,43 @@
...
+#
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema.define(version: 20150707225218) do
+
+  create_table "albums", force: :cascade do |t|
+    t.string   "title"
+    t.string   "author"
+    t.string   "description"
+    t.integer  "ranking",     default: 1
```
kariabancroft, line 20: Nice job setting this with a default value, although, should it be zero?

spec/controllers/albums_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    it "responds successfully with an HTTP 200 status code" do
+      get :index
+      expect(response).to be_success
+      expect(response).to have_http_status(200)
+    end
+  end
+# POST specs ---------------------------------------------------------------
+  describe "POST #create" do
+    context "Invalid album params" do
+      let(:album_params) do
```
kariabancroft, line 16: I might use a variable name with "invalid" to be more explicit about the types of tests that would be used with this

spec/controllers/albums_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      it "redirects to the album index page after saving" do
+        post :create, album_params
+        expect(response).to redirect_to(albums_path)
+      end
+    end
+ # PATCH specs --------------------------------------------------------------------
+    describe "PATCH #update" do
+      before(:each) do
+        @album = Album.new(title: "first title", author: "first author")
```
kariabancroft, line 60: You can use `Album.create` and get ride of the `save` line

spec/controllers/albums_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        @album.save
+      end
+
+      after(:each) do
+        @album.destroy
```
kariabancroft, line 65: Nice!

spec/controllers/albums_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
+require 'rails_helper'
+
+RSpec.describe AlbumsController, type: :controller do
```
kariabancroft, line 3: You can extract these tests to a shared example by using some `let` statements for the params and passing in the model type

spec/models/album_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'rails_helper'
+
+RSpec.describe Album, type: :model do
+  describe "model validations" do
+    it "requires a title" do
+      album = Album.new
+      expect(album).to_not be_valid
+      expect(album.errors.keys).to include(:title)
```
kariabancroft, line 8: Could you combine these both into a single test?

kariabancroft:
You did a nice job on this but I think you could clean up some of your controller and view functionality. The next thing that I would try if you want to refactor some would be to create some shared examples by using some additional `let` statements.

PR #25, opened 2015-08-02
Skill check: Refactoring

blip-lorist:
I realized this weekend that I had already done quite a bit of refactoring since my code review. Charles gave me some great DRY suggestions earlier. I'm very glad to have just the single Medium model and Media controller, because I've got less code to worry about refactoring. 

Here are the refactoring steps I took for this PR: 
- Preventing method chains: Created private set_media, set_movies, set_albums, and set_books methods that are called before actions.
- Updated custom routes so that they're more intuitive
- Cleaned up ugly link paths.

For the most part, I feel like my views only contains logic related to views (ie the conditional partials). But I'm not sure if that's the proper way to use partials, so I would love some feedback on that! Mahalo for your time.

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -1,37 +1,42 @@
 class MediaController < ApplicationController
+
+  before_action :set_medium, only: [:show, :upvote, :edit, :update, :destroy]
+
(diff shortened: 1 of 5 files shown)


kariabancroft:
Nice job! This looks good!

PR #26, opened 2015-08-02
MediaRanker -- Refactored

acmei:
This is my MediaRanker refactored. Still trying to understand how to best DRY out my controllers since they all have the same methods. User the ApplicationControler? Module? Super Controller?

All forms are in partials, the way this is done could be simpler but I'm not exactly sure how. All of my testing exists in shared examples.

Thanks!


kariabancroft:
Nice job! I like how cleaned up the shared error messages view is

PR #28, opened 2015-08-03
Refactoring Assignment

brittinator:
- This refactoring assignment was challenging because on this project I felt I had 'trimmed the fat' more so than on other projects, with rspec shared examples and partials. 
- I removed a lot of extra whitespace, and placed my hashes in the seed file so it's easier to see.
- I tried to make a url shortcut for the upvote, but then determined it was more work than necessary. 
- I also thought about putting the incrementing of the upvote into the models, but I'm using a rails method to change the rank by 1. By wrapping a method around it I'd actually be adding more code.
- I could've used the self.model probably in the controllers for the index action variable assignment, but I couldn't figure that out in the timebox I set for this assignment (3 hours).

### DIFF
--- a/app/controllers/albums_controller.rb
+++ b/app/controllers/albums_controller.rb
@@ -45,11 +45,11 @@ class AlbumsController < ApplicationController
     redirect_to :albums
   end
 
-# PRIVATE METHODS ----------------------------------------------------------
-private
   def self.model
     Album
   end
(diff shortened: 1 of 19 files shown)


kariabancroft:
Nice job on this! I think the only additional thing you could do is utilize polymorphic paths in your shared views to use the named paths rather than the URLs

PR #29, opened 2015-08-03
Refactored MediaRanker

a-lmx:
I used some of the code reviewer's suggestions to refactor Media Ranker. I moved view logic into the Application Helper when possible. I also got rid of the individual model views, rendering directly from the shared templates instead. Finally, I moved some of the generic actions into the Application Controller instead of having duplicate methods in each individual controller. I think I could do further refactoring and get all the methods into a single inherited controller given more time.

Though this refactor didn't save any lines of code, it leaves the logic more DRY, and in more appropriate locations.

:rabbit:


app/helpers/application_helper.rb
```diff
@@ -1,2 +1,54 @@
 module ApplicationHelper
+  def creator
+    case controller_name
+      when "albums"
```
kariabancroft, line 4: I'd update these case statements to use constants for the Albums, Books and Movies since it is the same string that is used for each

app/controllers/application_controller.rb
```diff
@@ -2,4 +2,57 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
+
+  def edit
```
kariabancroft, line 6: Nice!

app/views/albums/edit.html.erb
```diff
@@ -1,3 +0,0 @@
-<%= render partial: 'shared/form',  locals: { medium: @album,
```
kariabancroft, line 1: I like that you ended up skirting these views and accessing the shared views directly

kariabancroft:
Nice! I like what you ended up with here

PR #30, opened 2015-08-03
refactoring MediaRanker

sojeri:
I focused this refactoring mission on reducing repetition & moving methods to where they belonged (model vs helper vs controller vs view). One cool thing I did was more properly redo the URL helpers. Now in the views & controllers, it looks like I'm using rails generated URL helpers. It's only when you dig a little deeper into the helper files that you realize I've written my own.

I tried to work on chaining methods, too, but I think I'm still too biased about my own code. In the models, most of the chaining felt perfectly readable. When I wrote my scopes and methods in the beginning, it was partly with readability in mind. Something like `category.media.by_upvotes.only_ten` feels readable to me.
- I got stuck in `root.html.erb`, where I couldn't figure out how not to chain methods & how not to set a variable within a view without completely rewriting the underlying database structure. Is there a better way? ~~I feel like I've gone with six of one instead of half a dozen of the other.~~

Jeri


kariabancroft:
I like the way you wrote your own URL helpers! Per your PR comment, I don't see a ton of method chaining in `root.html.erb` - overall I think it looks good

PR #31, opened 2015-08-03
Media Ranker - Refactored

shannachau:
I really enjoyed this exercise. We learned some new things since the first time we did this, so it was nice to go back and implement some of them.

Here's a quick summary of some refactoring I did:
- Used before_action to set instance variables
- Added ranking_order scope to reduce chaining
- Consolidated more controller rspec tests into my shared examples!!!


kariabancroft:
You did a really nice job refactoring this! Controllers and specs look much more DRY now!

PR #32, opened 2015-08-03
Refactored Media Ranker

wangg131:
Refactoring Media Ranker

Some of the things that I changed were:
- Getting rid of assignment operator in the controllers. I made a migration to the schema so all of the model’s default ranking is 0 instead of having to declare model_name.rank = 0 in the controller’s upvote method. 
- Eliminating arithmetic operators in the controllers. I moved the upvote method of all three controllers to their respective models.
- I added before_action to DRY up the methods that required you to find the instance of each medium. `Ex: @movie = Movie.find(params[:id])`
- Really minor, but I used the ordered scope in the top scope. Scope on Scope action. Didn't save any lines but it seems cooler to look at so I went with it.
- There was a bunch of unnecessary logic for the params in the update methods. Since I now have a much better understanding of how params works, I was able to edit it and DRY it up.

We learned so much since completing Media Ranker, so it's cool to see how I can apply these new concepts to DRY up the code.

### DIFF
--- a/app/controllers/albums_controller.rb
+++ b/app/controllers/albums_controller.rb
@@ -37,8 +37,7 @@ class AlbumsController < ApplicationController
   end
 
   def upvote
-    @album.rank += 1
-    @album.save
+    Album.upvote(@album)
     redirect_to album_path(@album.id)
   end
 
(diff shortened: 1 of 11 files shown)


kariabancroft:
Nice! You did a great job refactoring! The scope within the scope is really smart!

PR #33, opened 2015-08-03
Bdl/master

Brennaleker:
refactoring


kariabancroft:
I like that you added the `locate_` methods, but it would be even more clean if you used a `before_action` to call these methods for each action that require it.

PR #34, opened 2015-08-03
Refactored MediaRanker

lindseyreno:
I added polymorphic paths to the views and the tests.  I also made the welcome index view a partial and made some other small changes to make everything consistent and dry up the code.


kariabancroft:
Excellent job with the polymorphic paths!

PR #35, opened 2015-08-03
Refactor of MediaRanker Project

sallyamoore:
For my refactor, I first incorporated my reviewers' comments by:
- extracting logic to find a specific record into a private method called in a before action
- removing some repetition
- assigning described_class.model to a variable in my controller specs
- added verification of the correct title to the spec for #create actions

I then pulled methods from my controller that 1) set the rank to 0 and 2) increased rank and put them in the models instead. This is repetitive, though, so I'm wondering how I'd DRY the code from here by creating a central place for methods like this. Same with the repeated before actions. Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,7 @@
 /tmp
 
 # Ignore SimpleCov files.
-coverage
+coverage/
 
 /model/.keep
 lib/assets/.keep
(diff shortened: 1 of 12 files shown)


kariabancroft:
Nice job implementing the `before_action`. I wonder if it would be easier to use `except` with the `before_action` rather than the `only` based on the number of action methods that require the find. Really nice overall

PR #36, opened 2015-08-03
Refactored Media Ranker

catchingash:


kariabancroft:
Nice job re-using a scope's logic as well as utilizing constants to prevent hard-coded strings and values

catchingash:
Thanks, @kariabancroft !

PR #37, opened 2015-08-03
Refactoring Media Ranker

QuartzBrandi:
Here's some refactoring to Media Ranker.
Unfortunately, some of the refactoring broke some of my rspec tests. (Specifically the last refactor of moving the rank assignment from the controller into the model.) I will have to look back at the rspec when I have time and try and fix them. At a glance, I'm not entirely sure why they broke.


app/controllers/movies_controller.rb
```diff
@@ -13,8 +13,8 @@ def new
   end
 
   def create
-    @movie = Movie.new(permit_params)
-    @movie.rank = 0
+    raise
```
kariabancroft: Ah! Don't check in your code with a `raise`

app/models/album.rb
```diff
@@ -5,7 +5,12 @@ class Album < ActiveRecord::Base
   # validates :description
   validates :rank, presence: true, numericality: { only_integer: true }
 
-  scope :best_first, -> (total) { all.order('rank DESC').limit(total) }
-  scope :best_first_all, -> { all.order('rank DESC') }
+  scope :best_first, -> (total) { order('rank DESC').limit(total) }
```
kariabancroft: You could also re-use the `best_first_all` scope within the `best_first` scope do you don't have to repeat the ordering logic

app/views/shared/_front_page_list.html.erb
```diff
@@ -1,20 +1,18 @@
...
+        <%= link_to medium.name, polymorphic_path(medium) %>
+        <small>
+          Ranked: <%= medium.rank %>
+        </small>
+      </li></h2>
+    <% end %>
+  </ol>
 
-    <%= link_to "View More #{category.capitalize}", "/#{category}" %>
+  <%= link_to "View More #{category.capitalize}", polymorphic_path(category) %>
```
kariabancroft, line 16: Nice job with the polymorphic_paths!

kariabancroft:
Nice job on this!

app/controllers/movies_controller.rb
QuartzBrandi: Whoops, I removed it in Atom, but forgot to Command+S.

PR #38, opened 2015-08-03
Refactor of media ranker, Katie

ktrops:


ktrops:
I moved strings into constance.
I added a before action hook :media to find the media.
I DRY'ed up the home page with partials.

kariabancroft:
Nice job! I wonder if there is a way to still list all of the media-type constants in only one place that would be accessible to all of your different controllers and models that require them.

PR #39, opened 2015-08-03
Refactor Pull Request

lilagrc:
To refactor this project, I used before actions to reduce reusing code in my controllers. I updated all url paths to use polymorphic paths in my partials, and I moved all of my rspec tests to shared examples.


app/controllers/albums_controller.rb
```diff
@@ -1,9 +1,14 @@
 class AlbumsController < ApplicationController
+  before_action :find_album, only: [:show, :edit, :update, :destroy]
 
   def self.model
     Album
   end
 
+  def find_album #Extracts out this search into one method
```
kariabancroft, line 8: This should be a private method since it's only used internally with the `before_action`s

app/controllers/albums_controller.rb
```diff
@@ -38,18 +42,16 @@ def create
   end
 
   def edit
```
kariabancroft, line 44: Take a look at the style guide for the recommended syntax for a method with no body

kariabancroft:
Nice job with your shared examples and polymorphic paths!

PR #40, opened 2015-08-03
Refactored Pull Request | Media Ranker

lo-k:
Here is my refactored version of Media Ranker. I was able to make all of the changes that @kariabancroft suggested in her code review except for successfully implementing a shared example using factories.

Highlights of this Refactoring:
1. Created shared example specs for book/album/movie model specs
2. Upgraded to using FactoryGirl to create test objects in show, edit and delete controller specs.
3. Upgraded to using FactoryGirl to modify object attributes in create and update controller specs
4. Added another spec to test the :best scope for each model
5. Wrote a private method to search for the album/book/movie needed in various controller specs. Added a before_action to call this private method before select methods


app/models/album.rb
```diff
@@ -1,9 +1,14 @@
...
   scope :best, -> (total) { order('rank DESC').limit(total)}
   
   # Validations
   validates :name, presence: true
   validates :description, presence: true
+
+  after_initialize :default_rank
+
+  def default_rank
+    self.rank ||= 0
```
kariabancroft, line 12: Nice use of the `||=`

kariabancroft:
Nice job! I like your private method that you added to use in the `before_action`!

PR #41, opened 2015-08-03
PR - refactor(wip)

EliaMG:
I got in over my head with this refactor. I was able to learn something new with having an STI table (media) and having all the models inherit from one model and use the same db table, so that worked well.

When I tried to refactor to have all my controllers inherit from ApplicationController, I got in over my head on inheritance and am too tired to think straight about it right now.

I will ask for help from a strong RoR TA to figure out what will work of my 2nd refactor, as at this point I have just broken my code. 

I think I am on the right track, but it is definitely broken code at the moment.

I had already used model methods to do the upvote functionality and all of my views were already partials, so I was trying to do a refactor that really pushed me beyond my comfort knowledge zone.


EliaMG:
Kept working on it because how broken it was make me not be able to sleep. Less DRY than I want, but am sharing some things in Application Controller and have gotten back to passing specs.

PR #42, opened 2015-08-03
Media Ranker refactor skill check

ElsaTKO:
Moved some logic into the model, added pluralize, added another before_action to set the media creator.

### DIFF
--- a/app/controllers/albums_controller.rb
+++ b/app/controllers/albums_controller.rb
@@ -3,6 +3,7 @@ class AlbumsController < ApplicationController
   before_action :find_and_set_media, only: [:upvote, :show, :edit, :update, :destroy]
 
   def index
+    # MOVE TO MODEL SCOPE
     @media = Album.all.order(rank: :desc)
     @new_media = Album.new
     @add_media = "an Album"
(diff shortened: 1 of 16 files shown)


app/controllers/albums_controller.rb
```diff
@@ -1,9 +1,10 @@
 class AlbumsController < ApplicationController
 
   before_action :find_and_set_media, only: [:upvote, :show, :edit, :update, :destroy]
+  before_action :set_creator, only: [:show, :new, :create, :edit, :update]
```
kariabancroft, line 4: Would it be less work to use an `except` rather than `only`?

app/models/album.rb
```diff
@@ -2,5 +2,9 @@ class Album < ActiveRecord::Base
   validates :name, presence: true
   validates :rank, numericality: { only_integer: true }
 
-  scope :top, -> { order(rank: :desc).limit(3) }
+  scope :ranked, -> { order(rank: :desc) }
+
+  def self.top(top_limit)
+    order(rank: :desc).limit(top_limit)
```
kariabancroft, line 8: Could you potentially re-use the `ranked` scope here for the `rank: :desc` logic?

kariabancroft:
Nice job!

PR #43, opened 2015-08-03
Media Ranker refactored code

MisShellyMac:
Here is my refactored code. Things I tried to focus on: using _path file names for all pages, correcting spacing and misaligned indentations, creating a method to dry up code to call in other methods, changing param name at bottom so it wasn't name create, adding .keep to gitignore file. I had planned to try to implement factory girl into spec but I ran out of time. I would still like to do that at some point. Thanks,
-Michelle

### DIFF
--- a/app/controllers/albums_controller.rb
+++ b/app/controllers/albums_controller.rb
@@ -1,12 +1,11 @@
 class AlbumsController < ApplicationController
 
-
     def index
       @albums = Album.all
     end
 
     def show
-      @album = Album.find(params[:id])
+      find_album
     end
 
     def new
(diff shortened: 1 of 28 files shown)


app/controllers/albums_controller.rb
```diff
@@ -16,46 +15,50 @@ def new
...
       @albums = Album.all
-      @new = Album.new(create_params[:album])
+      @new = Album.new(album_params[:album])
       @new.save
 
       render :index
     end
 
-    def edit
+    def find_album
```
kariabancroft, line 24: This should be a private method since it is used only as helpers for this controller.

app/controllers/albums_controller.rb
```diff
@@ -1,12 +1,11 @@
 class AlbumsController < ApplicationController
 
-
     def index
       @albums = Album.all
     end
 
     def show
-      @album = Album.find(params[:id])
+      find_album
```
kariabancroft, line 8: All `find_album` method calls could be moved into a `before_action`

app/controllers/albums_controller.rb
```diff
@@ -16,46 +15,50 @@ def new
...
     def update
-      @album = Album.find(params[:id])
-      album_params = create_params[:album]
-      @album.update(album_params)
+      find_album
+      params = album_params[:album]
+      @album.update(params)
       @album.save
-      redirect_to "/albums/#{@album.id}"
+      redirect_to album_path
```
kariabancroft, line 38: Do you still need to pass an `@album` object for this path?

app/views/movies/_index.html.erb
```diff
@@ -26,7 +26,7 @@
           <%= button_to "   👍+1   ", "movies/upvote/#{movie.id}", method: :patch %>
         </span>
         <span style="float:left; margin:5px">
-          <%= button_to "   👎-1   ", "moovies/downvote/#{movie.id}", method: :patch %>
+          <%= button_to "   👎-1   ", "movies/downvote/#{movie.id}", method: :patch %>
```
kariabancroft, line 29: You could potentially still create a custom route path for this

kariabancroft:
Nice job on this! You should still take your controllers to the next level by using a `before_action`. Using the named routes looks much better.

PR #44, opened 2015-08-03
Refactored MediaRanker

corinnepingul:
List of what I targeted to refactor:
1. albums controllers#create — figure out how to set the default rank in the model/db
2. model scope — call it in the controller as Album.rank_order, not Album.all.rank_order
3. before actions for controllers
4. don’t save on update methods
5. take out t/f in controllers for .nil?
6. shared/_home_display.html.erb
7. welcome/index.html.erb - figure out a way to not pass in the variable @media_types


kariabancroft:
Nice job!

PR #45, opened 2015-08-03
Refactoring hw - Media Ranker

camarleigh:
I'm super grateful this was assigned as hw. I like that it gives re-factoring a priority among all the new things we're doing.

m

### DIFF
--- a/app/controllers/albums_controller.rb
+++ b/app/controllers/albums_controller.rb
@@ -1,12 +1,17 @@
 class AlbumsController < ApplicationController
 
+  before_action :select_media, only: [:show, :edit, :update, :yes_vote, :destroy]
+
+  def select_media
+    @album = Album.find(params[:id])
+  end
+
   def index
     @albums = Album.best(5)
   end
 
   def create
-    @album = Album.new(permit_params[:album])
-    @album.save
+    @album = Album.create(album_params)
 
     redirect_to(albums_path)
   end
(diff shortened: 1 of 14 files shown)


app/controllers/albums_controller.rb
```diff
@@ -1,56 +1,60 @@
 class AlbumsController < ApplicationController
 
+  before_action :select_media, only: [:show, :edit, :update, :yes_vote, :destroy]
+
+  def select_media
```
kariabancroft, line 5: Nice job with this but you should move this to the private method section

app/controllers/books_controller.rb
```diff
@@ -1,57 +1,57 @@
...
   def index
     @books = Book.best(5)
   end
 
   def create
-    @book = Book.new(permit_params[:book])
-    @book.save
-
-    redirect_to(books_path)
+    @book = Book.create(book_params)
```
kariabancroft, line 14: The `create` method does the save operation for you, so if you still want to use the `save` logic below, you should continue using the `new` like you did in your previous commit

app/controllers/books_controller.rb
```diff
@@ -1,57 +1,57 @@
...
-    render :new
   end
 
   def show
-    @book = Book.find(params[:id])
   end
 
-  def edit #GET one book to modify.
-   @book = Book.find(params[:id])   
+  def edit #GET one book to modify.
```
kariabancroft, line 30: See the style guide for recommendations on what to do with methods with no body

app/controllers/movies_controller.rb
```diff
@@ -1,58 +1,59 @@
 class MoviesController < ApplicationController
  
+  before_action :select_media, only: [:show, :edit, :update, :yes_vote, :destroy]
+
+  def select_media
+    @movie = Movie.find(params[:id])
+  end
+  
   def index
     @movies = Movie.best(5)
```
kariabancroft, line 10: If you put the `5` as a constant within the `ApplicationController` you could update the # of records displayed on the index for ALL entries in one place rather than three (if that is potentially desirable)

app/views/albums/index.html.erb
```diff
@@ -12,7 +12,7 @@
       <tr>
         <td>Ranked:<%= album.vote %></td>
         <td><%= link_to "#{album.name}", album_path(album.id) %></td>
-        <td><%= button_to "Upvote", "/albums/#{album.id}/yes_vote", method: :patch, :class => "btn btn-default" %></td>
+        <td><%= button_to "Upvote", yes_vote_album_path(album.id), method: :patch, :class => "btn btn-default" %></td>
```
kariabancroft, line 15: Nice! Is this a route that already existed?

app/views/shared/_editcreate_form.html.erb
```diff
@@ -0,0 +1,2 @@
+
+<%= render partial: 'shared/editcreate_form', locals: { model: @album } %>
```
kariabancroft, line 2: It seems odd that this one in the "shared" folder is specifying the model as the `@album`

app/views/shared/_editcreate_album__forms.html.erb
```diff
@@ -0,0 +1,18 @@
+<%= form_for @album do |f| %>
```
kariabancroft, line 1: It still seems like this form is awfully similar to the _editcreate_book view

kariabancroft:
Excellent job cleaning up the controllers! I think there is still some additional work you should do to clarify the shared views.

PR #46, opened 2015-08-04
troubleshooting shared_example specs

knguyen0710:
I started by refactoring things that @kariabancroft suggested and when I got through with a good chunk and went to test it, my site wouldn't run. I started on shared examples but I'm not sure I'm doing those correctly either. Feedback on how on the index_controller_spec would be great.


Hamled:
I was able to clone your git repo on commit b09f0b3 (the latest from this pull request) and after doing `rake db:migrate` and `rake db:seed` it appears to run fine with the local development server.

What specific issues were you having after doing the refactoring?

Hamled:
I also looked into the shared examples you've setup for the controller specs. It looks like what you need at the top of each of those files is:

``` ruby
require 'rails_helper'
require 'support/index_controller_spec'
```

That, plus fixing an issue of an extra `end` line in the [Movies model spec file](https://github.com/knguyen0710/C3Projects--MediaRanker/blob/b09f0b3a2e0cace779bf916a915f45dbdebf8b86/spec/models/movie_spec.rb), was enough to get the tests to run (although not all pass).

